PR #5, opened 2020-08-28
Auto-convert TimeTypes to/from arrow time types

quinnj:
These are pretty cheap conversions and makes the Date, Time, Timestamp,
and Duration arrow types must more friendly to work with. I'm not sure
Interval has a real clear mapping, so it's left out for now (it also
mentions that Interval isn't even required for full arrow
compatibility). The last arrow type that doesn't currently map to a
native Julia type is Decimal. That will take a bit more digging because
the spec isn't very detailed on it's exact representation. I plan on
asking the arrow mailing list for more details and we can do that later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ julia:
   - 1.0
   - 1
   - nightly
+branches:
+  only:
+  - master
+  - gh-pages # For building documentation
+  - /^testing-.*$/ # testing branches
+  - /^v[0-9]+\.[0-9]+\.[0-9]+$/ # version tags
 matrix:
   allow_failures:
     - julia: nightly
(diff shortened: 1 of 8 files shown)


quinnj:
Whoop whoop! @bkamins @ExpandingMan, got tests passing on all platforms, and 32 + 64 bit. Things left on my TODO:
* test `Map`; it's not very well supported by other language implementations, but it's pretty much done, just need to test
* finish file support; we just need to write out the Footer flatbuffer thing and close the writing w/ the magic bytes again

After that, I'll probably start looking at doing the integration tests they propose in the apache arrow repo; it involves parsing json-ified blobs of arrow format and reading/writing with those.

Otherwise, we really just need to get people to try things out and report issues they find.

The other idea I had that would be interesting is building a kind of testing fuzzer; i.e. you could auto-generate various combinations of Union/List/Struct/Map/Primitive type combinations and just test round-tripping them. Right now the testing isn't super comprehensive wrt deep nesting.

PR #12, opened 2020-09-02
Check if column implements DataAPI.refarray for dict encoding

### LINKED ISSUE #2: CategoricalArrays strikes again
I think we should handle this:
```
julia> df = DataFrame(x6=categorical(["a","b"]), x7=categorical(["a",missing]))
2×2 DataFrame
│ Row │ x6   │ x7      │
│     │ Cat… │ Cat…?   │
├─────┼──────┼─────────┤
│ 1   │ a    │ a       │
│ 2   │ b    │ missing │

julia> Arrow.write("test.arrow", df)
ERROR: MethodError: no method matching arrowtype(::Arrow.FlatBuffers.Builder, ::Type{CategoricalValue{String,UInt32}})
```

quinnj:
Fixes #2. The `DataAPI.refarray` and `DataAPI.refpool` are meant to
allow custom array types to signal that they internally store integer
ref indices along with a value pool. This is a great integration with
the arrow format, which has a specific "dict encoded" format that
follows this same format (integer indices w/ value pool). So here we
leverage that by checking if an incoming column implements the interface
and, if so, signal that it should be dict encoded in the arrow format.

### DIFF
--- a/src/write.jl
+++ b/src/write.jl
@@ -473,7 +473,7 @@ function makenodesbuffers!(colidx, types, columns, dictencodings, fieldnodes, fi
             T = eltype(refvals)
             col = (x - one(T) for x in refvals)
         else
-            _, T, vals = msg.dictencodings[colidx]
+            _, T, vals = dictencodings[colidx]
             col = DictEncoder(col, vals, T)
         end
         len = _length(col)
--- a/src/write.jl
+++ b/src/write.jl
@@ -290,7 +290,6 @@ function Base.write(io::IO, msg::Message, blocks, sch)
             # @show typeof(col), col
             if msg.dictencodings !== nothing && haskey(msg.dictencodings, i)
                 refvals = DataAPI.refarray(col.data)
-                @show typeof(refvals), typeof(col.data)
                 if refvals !== col.data
                     T = eltype(refvals)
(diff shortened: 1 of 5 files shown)


src/write.jl
```diff
@@ -64,11 +64,18 @@ end
                 if col isa DictEncode
                     id = dictid[]
                     dictid[] += 1
-                    values = unique(col)
-                    dictencodings[i] = (id, encodingtype(length(values)), values)
+                    refpool = DataAPI.refpool(col.data)
+                    if refpool !== nothing
+                        values = refpool
+                        IT = eltype(DataAPI.refarray(col.data))
```
nalimilan, line 70: Note that `?DataAPI.refpool` currently doesn't guaranteed that you can call `eltype` on the result. The only guarantee is that "`refpool(A)[refarray(A)[I...]]` must be equal to `A[I...]`." Cf. https://github.com/JuliaData/DataAPI.jl/pull/1/#issuecomment-626141115. Maybe it's time to make the docstring more specific?

Specifically, for CategoricalArrays, it would be quite easy to return a `CategoricalPool` that supports `getindex`, but nothing else. But that wouldn't be very useful and would break this PR. :-p

PR #44, opened 2020-10-22
Prevent infinite loop while writing on CategoricalArrays

quinnj:
Supporting CategoricalArrays is a bit tricky. Ideally, we could use the
same interface as PooledArrays and just rely on `DataAPI.refarray` and
`DataAPI.refpool`, but alas, a `CategoricalArray` returns a
`CategoricalRefPool` from `DataAPI.refpool`, with `CategoricalValue`
elements. The core issue comes when we try to serialize the pool, which
is a recursive process: recursive until we reach a known "leaf" type, at
which point the recursion stops. Unfortunately, a `CategoricalValue`
isn't a known leaf type, so it's treated as a `StructType`, where each
of its fields are serialized. One of the fields is the
`CategoricalRefPool`, so we get stuck in a never-ending recursive loop
serializing `CategoricalValue`s and `CategoricalRefPool`s.

This PR proposes a quick hack where we check the `DataType` name for
`:CategoricalRefPool` or `:CategoricalArray` and if so, just unwrap the
values so the recursion will be broken. It's obviously a little hacky,
but also avoids taking on the always-problematic CategoricalArrays
dependency.

### DIFF
--- a/src/arraytypes/dictencoding.jl
+++ b/src/arraytypes/dictencoding.jl
@@ -94,7 +94,12 @@ function arrowvector(::DictEncodedType, x, i, nl, fi, de, ded, meta; dictencode:
         for i = 1:length(inds)
             @inbounds inds[i] -= 1
         end
-        data = arrowvector(DataAPI.refpool(x), i, nl, fi, de, ded, nothing; dictencode=dictencodenested, dictencodenested=dictencodenested, dictencoding=true, kw...)
+        pool = DataAPI.refpool(x)
+        # horrible hack? yes. better than taking CategoricalArrays dependency? also yes.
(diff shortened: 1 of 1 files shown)


quinnj:
@nalimilan @bkamins, thoughts here? It seems like the easiest path forward.

nalimilan:
The hack is fine with me as far as I'm concerned, I can't comment on whether there would be a better implementation in Arrow.jl.

>  One of the fields is the
> `CategoricalRefPool`, so we get stuck in a never-ending recursive loop
> serializing `CategoricalValue`s and `CategoricalRefPool`s.

FWIW the recursive loop is with `CategoricalPool`, not `CategoricalRefPool`.

Issue #84, opened 2020-12-22
writing column with missing / struct data errors


ericphanson:
```julia
using Arrow, Dates

struct MyDate
    date::Date
end
Arrow.ArrowTypes.registertype!(MyDate, MyDate)
table = (; date = [rand(Bool) ? MyDate(now()) : missing for _ = 1:10])
io = IOBuffer()
Arrow.write(io, table)
```

yields

```julia
julia> Arrow.write(io, table)
ERROR: TaskFailedException:MethodError: Cannot `convert` an object of type Day to an object of type Arrow.Date{Arrow.Flatbuf.DateUnitModule.DAY,Int32}
Closest candidates are:  convert(::Type{Arrow.Date{Arrow.Flatbuf.DateUnitModule.DAY,Int32}}, ::Date) at /home/ec2-user/.julia/packages/Arrow/CyJ4L/src/eltypes.jl:188
  convert(::Type{T}, ::T) where T at essentials.jl:171  Arrow.Date{Arrow.Flatbuf.DateUnitModule.DAY,Int32}(::Any) where {U, T} at /home/ec2-user/.julia/packages/Arrow/CyJ4L/src/eltypes.jl:165
Stacktrace: [1] arrowconvert(::Type{T} where T, ::Day) at /home/ec2-user/.julia/packages/Arrow/CyJ4L/src/arrowtypes.jl:32 [2] getindex at /home/ec2-user/.julia/packages/Arrow/CyJ4L/src/utils.jl:118 [inlined]
 [3] iterate at ./abstractarray.jl:986 [inlined] [4] writearray(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Type{Arrow.Date{Arrow.Flatbuf.DateUnitModule.DAY,Int32}}, ::Arrow.Converter{Arrow.Date{Arrow.Flatbuf.DateUnitModule.DAY,Int32},Arrow.ToStruct{Date,1,Array{Union{Missing, MyDate},1}}}) at /home/ec2-user/.julia/packages/Arrow/CyJ4L/src/utils.jl:50
[...]

PR #86, opened 2020-12-23
add `ArrowTypes.default` methods and tests for dates

ericphanson:
Fixes https://github.com/JuliaData/Arrow.jl/issues/84

The issue is that e.g. `zero(Date) == Day(0)` which is not a `Date`, so the `ArrowTypes.default` fallback gives the wrong type here. So I added fallbacks and tests.

I didn't really get how the tests are organized so I'm not sure I've added the tests in the right spot.


quinnj (approved):
Thanks! Great fix and thanks for adding the tests, they look great.

Issue #88, opened 2020-12-23
reconsidering the current type registration/serialization mechanism (and its internal usage)


ericphanson:
E.g.
```julia
julia> using Arrow, UUIDs                           
                                                                             
julia> table = (;x = [uuid4() for _ = 1:5])                                                                               
(x = UUID[UUID("03fec02f-3d33-4cf8-88e0-1e152faccff7"), UUID("a789baa7-9059-4bc4-b0d9-2af1dca4bf88"), UUID("2a4ecfd7-049e-
468d-8fed-8f967b732ee0"), UUID("91b9cad6-b84c-4ee8-aeaf-7448778fb1d5"), UUID("39edead7-1670-430a-8e4d-fefe3615c2d0")],)   

julia> Arrow.write("uuids.arrow", table)                                                                                  
"uuids.arrow"                                                                                                             
                                                                                                                          
julia> Arrow.Table("uuids.arrow")
Arrow.Table: (x = UUID[UUID("03fec02f-3d33-4cf8-88e0-1e152faccff7"), UUID("a789baa7-9059-4bc4-b0d9-2af1dca4bf88"), UUID("2
a4ecfd7-049e-468d-8fed-8f967b732ee0"), UUID("91b9cad6-b84c-4ee8-aeaf-7448778fb1d5"), UUID("39edead7-1670-430a-8e4d-fefe361
5c2d0")],)
```                                

In a new session:
```julia
julia> using Arrow, UUIDs

julia> Arrow.Table("uuids.arrow")
┌ Warning: unsupported ARROW:extension:name type: "JuliaLang.UUID"
└ @ Arrow.ArrowTypes ~/.julia/packages/Arrow/CyJ4L/src/arrowtypes.jl:141
Arrow.Table: (x = NamedTuple{(:value,),Tuple{UInt128}}[(value = 0x03fec02f3d334cf888e01e152faccff7,), (value = 0xa789baa790594bc4b0d92af1dca4bf88,), (value = 0x2a4ecfd7049e468d8fed8f967b732ee0,), (value = 0x91b9cad6b84c4ee8aeaf7448778fb1d5,), (value = 0x39edead71670430a8e4dfefe3615c2d0,)],)
```

I believe this is due to the write-time registration of types at

https://github.com/JuliaData/Arrow.jl/blob/3ab2b18829c1656198a85759360389b6bbb22ab3/src/arraytypes/struct.jl#L86

ericphanson:
[...]

PR #89, opened 2020-12-26
add default UUID <-> UInt128 Arrow type mapping

jrevels:
(vaguely related: https://issues.apache.org/jira/browse/ARROW-2051)

Naively this seems like a more reasonable Julia-agnostic default than the fallback that `UUID` currently hits (an example of which can be found in #88). My use case is that I'm spec'ing out a Julia-agnostic, Arrow-based format and I'm specifying that "UUID columns are stored as 128-bit binary values" - which seems reasonable but technically conflicts with Arrow.jl's current `UUID` fallback.

If this isn't a good idea to include by default here, I can put these definitions in my downstream package - just seemed like a good idea to upstream it if possible to avoid potential conflicts w/ other downstream packages that might rely on the existing default behavior

@quinnj if you're good with this idea, I can add tests + take out of draft mode.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,6 +14,7 @@ PooledArrays = "2dfb63ee-cc39-5dd5-95bd-886bf059d720"
 SentinelArrays = "91c51154-3ec4-41a3-a24f-3f23e20d615c"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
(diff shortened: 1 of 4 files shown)


quinnj (approved):
Yeah, this looks great to me; let's add some tests and rock n' roll.

jrevels:
Ha, okay, adding a basic test here was easier than I had imagined it'd be 😁

AFAICT this is ready for merge pending @quinnj's review (let me know if there are any additional tests I should add!)

quinnj (approved):

Issue #87, opened 2020-12-23
downstream packages need to put `Arrow.ArrowTypes.registertype!` statements in `__init__`


ericphanson:
Otherwise `Arrow.ArrowTypes.ARROW_TO_JULIA_TYPE_MAPPING` doesn't get updated on package load, since the `registertype!` call gets executed at precompile time and that state is not preserved for using time. It seems like this is just a consequence of the design choice to use a global dictionary instead of dispatch (like `StructTypes` does). This caught me off-guard so I figured I'd file an issue; I think it should be mentioned in the docs at least.

PR #96, opened 2021-01-04
Add warning for `Arrow.ArrowTypes.registertype!`

ericphanson:
Closes https://github.com/JuliaData/Arrow.jl/issues/87. I *think* I got the [admonition](https://docs.julialang.org/en/v1/stdlib/Markdown/#Admonitions) syntax right, but I didn't test it locally.


quinnj (approved):
Thanks for doing this

PR #103, opened 2021-01-07
change UUID <-> Arrow mapping to (de)serialize to/from 16-byte FixedSizeBinary

jrevels:
...to make tables written out by Arrow.jl more portable. 

I'm dumb and didn't realize e.g. `pyarrow` doesn't support 128-bit ints, so currently UUID-containing tables will fail to be read by e.g. `pa.feather.read_feather(...)`.

There's even pyarrow documentation that [suggests to Python users that UUIDs can be written as FixedSizeBinary columns](https://arrow.apache.org/docs/python/extending_types.html), so this seems like a more friendly/correct way to do this.

### DIFF
--- a/docs/src/manual.md
+++ b/docs/src/manual.md
@@ -53,9 +53,11 @@ Apart from letting other packages have all the fun, an `Arrow.Table` itself can 
 
 In the arrow data format, specific logical types are supported, a list of which can be found [here](https://arrow.apache.org/docs/status.html#data-types). These include booleans, integers of various bit widths, floats, decimals, time types, and binary/string. While most of these map naturally to types builtin to Julia itself, there are a few cases where the definitions are slightly different, and in these cases, by default, they are converted to more "friendly" Julia types (this auto conversion can be avoided by passing `convert=false` to `Arrow.Table`, like `Arrow.Table(file; convert=false)`). Examples of arrow to julia type mappings include:
 
-* `Date`, `Time`, `Timestamp`, and `Duration` all have natural Julia defintions in `Dates.Date`, `Dates.Time`, `TimeZones.ZonedDateTime`, and `Dates.Period` subtypes, respectively. 
+* `Date`, `Time`, `Timestamp`, and `Duration` all have natural Julia defintions in `Dates.Date`, `Dates.Time`, `TimeZones.ZonedDateTime`, and `Dates.Period` subtypes, respectively.
 * `Char` and `Symbol` Julia types are mapped to arrow string types, with additional metadata of the original Julia type; this allows deserializing directly to `Char` and `Symbol` in Julia, while other language implementations will see these columns as just strings
+* Similarly to the above, the `UUID` Julia type is mapped to a 128-bit `FixedSizeBinary` arrow type.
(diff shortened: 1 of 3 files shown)


quinnj:
Failing tests?

jrevels:
Ah, was missing some methods for FixedSizeListType types it seems. Fixed IIUC.

It's a shame that this entails a significant perf hit for `copy`:

```
julia> @time copy(uuids_uint128);
  0.002593 seconds (2 allocations: 3.248 MiB)

julia> length(uuids_fixed)
212886

julia> typeof(uuids_fixed)
Arrow.FixedSizeList{Base.UUID,Array{UInt8,1}}

julia> length(uuids_uint128)
212886

julia> typeof(uuids_uint128)
Arrow.Primitive{Base.UUID,Array{UInt128,1}}

julia> @time copy(uuids_fixed);
  0.167757 seconds (1.28 M allocations: 94.203 MiB, 2.81% gc time)

julia> @time copy(uuids_uint128);
  0.001690 seconds (2 allocations: 3.248 MiB)
```

I imagine we could optimize further to lessen/remove the overhead, though I think it'd preferable to do that in a follow-up

jrevels:
Thanks to @SimonDanisch the UInt128 <-> NTuple{16,UInt8} cast employed here is now non-allocating 🔥

Reduces allocations for `copy(uuids_fixed)` a bit (see below) but looks like there's some additional performance still left on the table.

```
julia> @time copy(uuids_fixed);
  0.142450 seconds (1.06 M allocations: 74.713 MiB, 2.43% gc time)
```

I'll dig into it in a separate PR when I get a chance but AFAICT this is now good to go pending green CI.

jrevels:
@quinnj This is now ready for review, recommend looking at #104 at the same time (ref https://github.com/JuliaData/Arrow.jl/pull/104#issuecomment-757369874, which I can take care of once either is merged) 🙂 thanks!

quinnj (approved):

ericphanson:
BTW the PR title here says FixedSizeBinary, but what was implemented was FixedSizeList; why is that?

Asking since [arrow is considering a canonical extension for uuids that uses FixedSizeBinary](https://github.com/apache/arrow/pull/41299).

jrevels:
> BTW the PR title here says FixedSizeBinary, but what was implemented was FixedSizeList; why is that?

IIRC back when this PR was authored, Arrow.jl would (de)serialize Julia values with type `FixedSizeList` w/ `UInt8` eltype as Arrow `FixedSizeBinary` values, and for w/e reason, that seemed like the right/easiest path to use in this PR (can't really remember the rationale for that, or whether there even was one 😅)

idk if that is still true nowadays though

here's the relevant `arrowtype` definition that existed when this PR was authored:

```
function arrowtype(b, x::FixedSizeList{T, A}) where {T, A}
    N = ArrowTypes.getsize(Base.nonmissingtype(T))
    if eltype(A) == UInt8
        Meta.fixedSizeBinaryStart(b)
        Meta.fixedSizeBinaryAddByteWidth(b, Int32(N))
        return Meta.FixedSizeBinary, Meta.fixedSizeBinaryEnd(b), nothing
    else
        children = [fieldoffset(b, "", x.data)]
        Meta.fixedSizeListStart(b)
        Meta.fixedSizeListAddListSize(b, Int32(N))
        return Meta.FixedSizeList, Meta.fixedSizeListEnd(b), children
    end
end
```

PR #104, opened 2021-01-09
add isbitstype optimized path for FixedSizeList getindex

jrevels:
ref https://github.com/JuliaData/Arrow.jl/pull/103#issuecomment-757354386

#103 before this PR:

```
julia> @time copy(uuids_fixed);
  0.142450 seconds (1.06 M allocations: 74.713 MiB, 2.43% gc time)
```

#103 after this PR:

```
julia> @time copy(y.recording_uuid);
  0.016563 seconds (2 allocations: 3.248 MiB)
```

Order of magnitude improvement + allocations/memory now on par with pre-#103 behavior. Nice!

Still an order of magnitude of runtime perf on the table compared to the equivalent `Primitive` though. will play around and check whether branching changes/direct pointer manipulation might recover some of that.

### DIFF
--- a/src/arraytypes/fixedsizelist.jl
+++ b/src/arraytypes/fixedsizelist.jl
@@ -31,12 +31,19 @@ Base.size(l::FixedSizeList) = (l.ℓ,)
 
 @propagate_inbounds function Base.getindex(l::FixedSizeList{T}, i::Integer) where {T}
     @boundscheck checkbounds(l, i)
-    N = ArrowTypes.getsize(Base.nonmissingtype(T))
-    off = (i - 1) * N
-    if Base.nonmissingtype(T) !== T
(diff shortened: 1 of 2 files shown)


jrevels:
Using the same `_unsafe_cast` strategy from #103, we now get:

```
julia> @time copy(y.recording_uuid);
  0.009492 seconds (2 allocations: 3.248 MiB) 
```

Still slower than the `Primitive` direct copy but now within the same order of magnitude, so I'm happy 😁 

Once either #103 or this PR is merged, the remaining PR should be  refactored to use the same underlying `_unsafe_cast!` utility defined here.  (shouldn't matter which one is merged first)

quinnj (approved):
LGTM, let's just rebase the `_unsafe_cast` out since #103 is merged

jrevels:
> let's just rebase the _unsafe_cast out since #103 is merged

done!

Issue #113, opened 2021-01-27
Unsigned integers as indices in DictEncoded type


dmbates:
Per the discussion around https://julialang.slack.com/archives/C674VR0HH/p1611699007085600 the expectation is that unsigned integers in a `DataAPI.refarray` will be changed to signed integers when converting to `DictEncoded` but that doesn't seem to be happening.
```julia
julia> io = Arrow.write(IOBuffer(), (v = PooledArray(repeat(["a","b"], inner=5)),));

julia> tbl = Arrow.Table(seekstart(io))
Arrow.Table: (v = ["a", "a", "a", "a", "a", "b", "b", "b", "b", "b"],)

julia> typeof(tbl.v)
Arrow.DictEncoded{String, UInt8, Arrow.List{String, Int32, Vector{UInt8}}}
```

I notice that methods for `signedtype` are defined in src/arraytypes/dictencoding.jl but I don't seen any uses of that function.  It may be that this is vestigial and `encodingtype` is used instead.

dmbates:
If I had to guess I would say that line 121 of src/arraytypes/dictencoding.jl, which is currently
```
            inds = copy(DataAPI.refarray(x))
```
should coerce to `Vector{signedtype(eltype(DataAPI.refarray(x))}` instead.

quinnj:
Fixed in #119

Issue #112, opened 2021-01-27
Producing unsigned dict encoding indices; should be signed


quinnj:
From @dmbates on slack:

Yes, I read a CSV file then wrote an Arrow file and it ended up with UInt32 indices in the Arrow file.  I can read it using pyarrow.feather.read_table which reports
3:28
PyObject pyarrow.Table
lab: dictionary<values=string, indices=uint32, ordered=0> not null
subid: dictionary<values=string, indices=uint32, ordered=0> not null
trial_num: int64 not null
trial_type: dictionary<values=string, indices=uint32, ordered=0>
stimulus_num: int64
looking_time: double
trial_error: bool not null
trial_error_type: dictionary<values=string, indices=uint32, ordered=0>
method: dictionary<values=string, indices=uint32, ordered=0> not null
ra: dictionary<values=string, indices=uint32, ordered=0>

PR #114, opened 2021-01-27
Enforce signed indices on DictEncoded type

dmbates:
- Addresses #112 and #113
- test to be added

### DIFF
--- a/src/arraytypes/dictencoding.jl
+++ b/src/arraytypes/dictencoding.jl
@@ -113,12 +113,14 @@ function arrowvector(::DictEncodedType, x, i, nl, fi, de, ded, meta; dictencode:
     validity = ValidityBitmap(x)
     if !haskey(de, id)
         # dict encoding doesn't exist yet, so create for 1st time
+        ET = encodingtype(length(DataAPI.levels(x)))
         if DataAPI.refarray(x) === x
             # need to encode ourselves
-            x = PooledArray(x, encodingtype(length(x)))
+            x = PooledArray(x, ET)
             inds = DataAPI.refarray(x)
         else
-            inds = copy(DataAPI.refarray(x))
+            ra = DataAPI.refarray(x)
+            inds = copyto!(similar(ra, ET), ra)
         end
         # adjust to "offset" instead of index
         for i = 1:length(inds)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -216,6 +216,10 @@ x = PooledArray(["hey", missing])
 x2 = Arrow.toarrowvector(x)
 @test isequal(copy(x2), x)
 
+# signed indices for DictEncodedType #112 #113 #114
+av = Arrow.toarrowvector(PooledArray(repeat(["a", "b"], inner = 5)))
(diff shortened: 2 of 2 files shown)


quinnj (approved):
This is great, thanks. Would be great to have a test.

dmbates:
There are a couple of things that I am still unsure about.  I changed the calculation of the type of the indices to be based on the length of the levels of `x` in the `if !haskey(de, id)` branch rather than being based on `length(x)`.  There is a similar calculation for `ET` in the else branch but I'm not entirely sure what that branch is for and what `x` is in that branch.

Also, the `encodingtype` method divides the typemax of signed types by 2. I don't think that is necessary.  It is the unsigned types that have half the range when converted to signed.

src/arraytypes/dictencoding.jl
```diff
@@ -113,12 +113,14 @@ function arrowvector(::DictEncodedType, x, i, nl, fi, de, ded, meta; dictencode:
     validity = ValidityBitmap(x)
     if !haskey(de, id)
         # dict encoding doesn't exist yet, so create for 1st time
+        ET = encodingtype(length(DataAPI.levels(x)))
```
quinnj, line 116: I think we want to avoid calling `DataAPI.levels` here because we'll end up doing the "pooling" operation twice: once here, and again below when we call `PooledArray(x, ET)`. 

I think this would be my recommended approach:
```julia
# dict encoding doesn't exist yet, so create for 1st time
        if DataAPI.refarray(x) === x
            # need to encode ourselves
            x = PooledArray(x, encodingtype(length(x)))
            refa = DataAPI.refarray(x)
        else
            refa = copy(DataAPI.refarray(x))
        end
        inds = converter(encodingtype(length(refa)), refa)
```

dmbates: But the minimum size of the index type depends on the number of levels not the length of the vector.  For me a typical case for a DictEncoding can have a small number of levels, such as would be expressible by an Int8, but the size of the vector itself would require an Int32.  Demographic variables like Sex, Race, IncomeLevel, ... are typical cases for me.

May I suggest an alternative which is to add a named Bool argument `signed` to the `PooledArray` constructor which will compress the index size to signed integer types.  I made the suggestion assuming that PooledArrays implemented the `DataAPI` methods like `levels` so it would be easy to determine the number of levels in the call to `converter`.  But now I see it doesn't.  Is there a reason why not?

dmbates: Ah, I see now that it is `DataAPI.refpool` that is implemented in PooledArrays.

So my suggestion is either to change `PooledArrays` to use signed integers for the refarray by default (I think I may have been guilty of suggesting unsigned integers initially in case the size of the refpool was between 128 and 255) or allow for a `signed` named argument in the PooledArray constructor .

In either case, can we use `converter(encodingtype(length(DataAPI.refpool)), refa)` for the inds?

quinnj: Yes, `converter(encodingtype(length(DataAPI.refpool)), refa)` sounds great

PR #115, opened 2021-01-28
Another attempt at signed indices

dmbates:


dmbates:
This currently fails tests.  I think it is because I don't understand how converter works.  I know it is designed to be lazy but it seems that the evaluation is not taking place.

quinnj:
Oh darn, that's because we have `indices::Vector{S}`, so it's expecting the indices to be an actual `Array` instead of some AbstractArray. Hmmm, what can we do here.

quinnj:
Thinking this over some more, I think we can still all the pieces we want. I think we move `pool = DataAPI.refpool(x)` up higher, then call `x = PooledArray(x, encodingtype(length(pool)))` in the first branch, and your `copyto!` w/ `similar` in the 2nd path. Does that sound ok?

dmbates:
I'm getting confused.  If you move the call to `DataAPI.refpool` higher isn't there a risk that there will be no specific method for `typeof(x)` and it will fall back on the default method that returns `nothing`?

dmbates:
An alternative may be to use `convert(Vector{encodingtype(length(pool))}, refa)` instead converter.  `convert` returns the original vector if it is of the desired type.

It seem to me that either we will repeat the action of creating the pool to get the right size of integer in the call to `PooledArray` or we will potentially assign the wrong size of integers in the call to `PooledArray` and need to convert later.  That is why I suggested changing the `PooledArray` constructor so that it could use signed integers.  The information to get the right size is in the `PooledArray` constructor - it is just that right now it will always use unsigned integers for the refarray.

quinnj:
> I'm getting confused. If you move the call to DataAPI.refpool higher isn't there a risk that there will be no specific method for typeof(x) and it will fall back on the default method that returns nothing?

Yes, you're right; sorry, I'm not being very helpful with suggestions here.

Yeah, I think we probably need to add the `signed` keyword to PooledArrays, which wouldnt' be too hard. That solves most of the problems here. I can do that later tonight unless you want to do it sooner.

dmbates:
Great.  Thanks for volunteering,

Issue #116, opened 2021-01-29
DictEncoded doesn't write as DictEncoded


dmbates:
`Arrow.toarrowvector` of a Arrow.DictEncoded type returns an `Arrow.List`.
```
julia> atbl = Arrow.Table(fnm);

julia> typeof(atbl.a)
Arrow.DictEncoded{String, UInt8, Arrow.List{String, Int32, Vector{UInt8}}}

julia> Arrow.toarrowvector(atbl.a)
6-element Arrow.List{String, Int32, Arrow.ToList{UInt8, true, String, Int32}}:
 "a"
 "a"
 "a"
 "b"
 "b"
 "b"
```
Perhaps it doesn't detect that `Arrow.DictEncoded` is a factor-like (in the R sense) object.

I can look at this if you wish but I'm not entirely sure where

quinnj:
Fixed in #119

PR #119, opened 2021-01-30
Rework dict encoding of PooledArray/CategoricalArray to fix outstandi…

### LINKED ISSUE #117: Errors writing file with missing in categorical
```
julia> using Base.Filesystem, CategoricalArrays, DataFrames
       open(Filesystem.tempname(), "w") do f
           Arrow.write(f, DataFrame(x=CategoricalArray(["a", "bb", missing])))
       end
ERROR: InexactError: trunc(UInt32, -1)
```

Sometimes I get a different error on another dataframe but I don't have a reproducible example for this one:

```
julia> open(path, "w") do f
           Arrow.write(f, df)
       end
ERROR: BoundsError: attempt to access 52-element CategoricalArrays.CategoricalRefPool{Union{Missing, CategoricalValue{String,UInt32}},CategoricalPool{String,UInt32,CategoricalValue{String,UInt32}}} with indices 0:51 at index [52]
Stacktrace:
 [1] throw_boundserror(::CategoricalArrays.CategoricalRefPool{Union{Missing, CategoricalValue{String,UInt32}},CategoricalPool{String,UInt32,CategoricalValue{String,UInt32}}}, ::Tuple{Int64}) at ./abstractarray.jl:541
 [2] checkbounds at ./abstractarray.jl:506 [inlined]
 [3] getindex(::CategoricalArrays.CategoricalRefPool{Union{Missing, CategoricalValue{String,UInt32}},CategoricalPool{String,UInt32,CategoricalValue{String,UInt32}}}, ::Int64)```
```

I don't know if this usage is supported but if not the error message could say so.

[69666777] Arrow v1.2.1
Julia 1.5.3

quinnj:
…ng issues

Fixes #117, #116, and #113. For #116, we just need to special case if user happens to pass in a DictEncoded themselves. We need to pass it through to the `toarrowvector` method that no-ops. For #113, we require the new functionality in PooledArrays that allows passing the `signed` and `compress` keyword arguments to ensure we get signed refs for our dict encoding. For #117, we add CategoricalArrays as a test dependency and ensure that if it contains any `missing` value, we *don't* recode the indices values down by 1, since the `missing` ref is 0, so other refs can already be considered "offsets". If there are no `missing`, then we still need to recode down since refs should always start from 0 in arrow format.

cc: @dmbates on the #113 portion (PR currently fails because we're tagging new PooledArrays functionality first)

cc: @nalimilan, I think the logic here for CategoricalArrays is better, but I wouldn't mind a brief review from you


src/arraytypes/dictencoding.jl
```diff
@@ -114,19 +114,29 @@ function arrowvector(::DictEncodedType, x, i, nl, fi, de, ded, meta; dictencode:
         # dict encoding doesn't exist yet, so create for 1st time
         if DataAPI.refarray(x) === x
```
nalimilan, line 115: Note that there's no guaranty that `refarray` returns an array of contiguous integer indices, nor that `refpool` is defined when `refarray` is. To make this more general it would make sense to check that, and if that's not the case then create a `PooledArray` as in the first branch.

src/arraytypes/dictencoding.jl
```diff
@@ -114,19 +114,29 @@ function arrowvector(::DictEncodedType, x, i, nl, fi, de, ded, meta; dictencode:
...
-            @inbounds inds[i] -= 1
-        end
-        pool = DataAPI.refpool(x)
         # horrible hack? yes. better than taking CategoricalArrays dependency? also yes.
         if typeof(pool).name.name == :CategoricalRefPool
-            pool = [get(pool[i]) for i = 1:length(pool)]
+            if eltype(x) >: Missing
+                pool = vcat(missing, pool.pool.levels)
+            else
+                pool = pool.pool.levels
```
nalimilan, line 130: This would avoid relying on internal fields:
```suggestion
            if eltype(x) >: Missing
                pool = vcat(missing, DataAPI.levels(x))
            else
                pool = DataAPI.levels(x)
```

src/arraytypes/dictencoding.jl
quinnj: I don't think we're assuming that `refarray` returns contiguous integers, because we're still doing `inds = copyto!(similar(Vector{signed(eltype(refa))}, length(refa)), refa)`, which should still work if non-contiguous. 

I guess we could add the check `|| refpool(x) === nothing` here. Curious, what are cases where `refarray` is defined, but not `refpool`?

nalimilan: > I don't think we're assuming that `refarray` returns contiguous integers, because we're still doing `inds = copyto!(similar(Vector{signed(eltype(refa))}, length(refa)), refa)`, which should still work if non-contiguous.

At least you assume `signed(eltype(refa))` is defined, but the docstring for `refarray` doesn't say that. Currently I'm not aware of types for which that's not the case, but better be safe. (I have no idea about whether contiguity is required.)

> I guess we could add the check `|| refpool(x) === nothing` here. Curious, what are cases where `refarray` is defined, but not `refpool`?

Well again I'm just following the docstrings. Historically, we introduced `refarray` and `refvalue` before `refpool`, so clearly implementing the two former doesn't imply implementing the latter. But conversely, when you implement `refpool` you have to implement `refarray`, otherwise you can't satisfy requirements from its docstring. I think the asymmetry makes sense since in general, implementing `refpool` can be harder as it requires knowing the set of possible values in advance.

Issue #121, opened 2021-02-02
Unsafe conversion to signed integer types


dmbates:
at `src/arraytypes/dictencoding.jl:125` indices with an unsigned type are converted to the signed type of the same size, which can fail if, e.g., you try to convert `UInt8(130)` to `Int8`.  I have examples where that occurs.

I think the best fix is to use `length(pool)` to determine the signed type.  `pool` has already been evaluated.

I can submit a PR if you agree with this approach @quinnj

quinnj:
Yes, that sounds good to me.

PR #122, opened 2021-02-04
Use pool length in signed int conversion

dmbates:
- addresses #121


dmbates:
The other methods for signedtype now are redundant, I think, but there is no harm in keeping them.

quinnj (approved):

quinnj:
Thanks!

PR #124, opened 2021-02-05
fix accidental invocation of _unsafe_load_tuple

jrevels:
This optimization was only meant to be used when the underlying data buffer was straight bytes. Looking at this, it might actually also be refactorable to be suitable in the more general case when `l.data isa Vector{D} && isbitstype(D)`, but AFAICT seems like that case wouldn't arise in practice that often (unless I'm mistaken).

Bumped the Project.toml to facilitate tagging after merging.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arrow"
 uuid = "69666777-d1a9-59fb-9406-91d4454c9d45"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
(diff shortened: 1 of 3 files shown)


quinnj (approved):
I'll admit I didn't fully understand the original optimization here well enough; I agree that this is a good fix for now, but it would be nice to generalize the fastpath more when the type sizes match. It should be very straightforward to get pointers to Arrow.Primitive if needed to enable this.

PR #150, opened 2021-03-13
define caller-overloadable arrownameof(T) to allow callers to specify preferred Julia <-> Arrow type matching behavior

jrevels:
EDIT: changed the approach on this a bit due to https://github.com/JuliaData/Arrow.jl/pull/150#issuecomment-802150610

given:

```jl
julia> struct Baz{T}
           x::Type{T}
       end

julia> Arrow.ArrowTypes.registertype!(Baz, NamedTuple{(:x,),Tuple{String}})

julia> Arrow.ArrowTypes.arrowconvert(::Type{NamedTuple{(:x,),Tuple{String}}}, b::Baz) = (x=string(b.x),)

julia> Baz(s::String) = s === "Int64" ? Baz(Int) : Baz(DataType)
Baz
```

on `main`:

```jl
julia> t = Arrow.Table(Arrow.tobuffer((x=[Baz(Int)],)))
ERROR: ArgumentError: type does not have a definite number of fields
Stacktrace:
 [1] fieldcount at ./reflection.jl:725 [inlined]
 [2] arrowvector(::Arrow.ArrowTypes.StructType, ::Arrow.ToStruct{Type{Int64},1,Array{Baz{Int64},1}}, ::Int64, ::Int64, ::Int64, ::Dict{Int64,Any}, ::Array{Arrow.DictEncoding,1}, ::Nothing; kw::Base.Iterators.Pairs{Symbol,Union{Nothing, Bool},NTuple{5,Symbol},NamedTuple{(:dictencode, :compression, :largelists, :denseunions, :dictencodenested),Tuple{Bool,Nothing,Bool,Bool,Bool}}}) at /Users/jarrettrevels/.julia/dev/Arrow/src/arraytypes/struct.jl:89
```

on this PR:

```jl
julia> Arrow.ArrowTypes.arrownameof(::Type{<:Baz}) = "JuliaLang.Baz"

julia> t = Arrow.Table(Arrow.tobuffer((x=[Baz(Int)],)))
Arrow.Table: (x = Baz[Baz{Int64}(Int64)],)
```

Additionally, this allows the caller to fix issues where the default behavior doesn't match the desired behavior:

```jl
julia> struct Foo{T}
           x::T
       end

julia> Arrow.ArrowTypes.registertype!(Foo, NamedTuple{(:x,),Tuple{String}})

julia> Arrow.ArrowTypes.arrowconvert(::Type{NamedTuple{(:x,),Tuple{String}}}, f::Foo) = (x=string(f.x),)

# The default behavior will miss the above overloads entirely since the type parameters are included in
# the name, but (perhaps unintuitively) will still work, just not in the way the caller intended
julia> t = Arrow.Table(Arrow.tobuffer((x=[Foo(1)],)))
Arrow.Table: (x = Foo{Int64}[Foo{Int64}(1)],)

# This PR allows the caller to add the "missing link" here to the desired matching behavior
julia> Arrow.ArrowTypes.arrownameof(::Type{<:Foo}) = "JuliaLang.Foo"

# and tada, now we get the desired behavior 
julia> t = Arrow.Table(Arrow.tobuffer((x=[Foo(1)],)))
Arrow.Table: (x = Foo[Foo{String}("1")],)
```


jrevels:
oof, the test failure here is basically due to the same issue brought up in https://github.com/JuliaData/Arrow.jl/issues/88#issuecomment-798806038. 

Basically we have a `Arrow.KeyValue{String,Union{Missing, Int32}}` type that hits the code path here: https://github.com/JuliaData/Arrow.jl/blob/3ab2b18829c1656198a85759360389b6bbb22ab3/src/arraytypes/struct.jl#L84

but due to a previous test case, the `Arrow.ArrowTypes.JULIA_TO_ARROW_TYPE_MAPPING` is:

```
julia> Arrow.ArrowTypes.JULIA_TO_ARROW_TYPE_MAPPING
Dict{Symbol,Tuple{String,Type}} with 9 entries:
  :WrappedZonedDateTime => ("JuliaLang.WrappedZonedDateTime", WrappedZonedDateTime)
  :Symbol               => ("JuliaLang.Symbol", String)
  :WrappedDate          => ("JuliaLang.WrappedDate", WrappedDate)
  :UUID                 => ("JuliaLang.UUID", NTuple{16,UInt8})
  :Char                 => ("JuliaLang.Char", UInt32)
  :WrappedDateTime      => ("JuliaLang.WrappedDateTime", WrappedDateTime)
  :CustomStruct         => ("JuliaLang.CustomStruct", CustomStruct)
  :KeyValue             => ("JuliaLang.Arrow.KeyValue{Int32,Float32}", KeyValue{Int32,Float32})
  :WrappedTime          => ("JuliaLang.WrappedTime", WrappedTime)
```

So `KeyValue{String,Union{Missing, Int32}}` here gets erroneously mapped to `KeyValue{Int32,Float32}`

Hmmm I could probably tweak this to elide type params/be more generic for `StructType` types but I wonder if this is a smell that the use case targeted by this PR is at odds with some expectations about how the registration mechanism is used internally? Maybe not though :p I'll play with it

jrevels:
Okay, switched tactics here. Instead of changing the built-in type mapping mechanism, I've made it more overloadable, keeping the existing behavior as the default fallback. See OP for an example.

This avoids the breakage demonstrated in https://github.com/JuliaData/Arrow.jl/pull/150#issuecomment-802150610, while also enabling the features this PR was originally seeking to enable

jrevels:
Okay, added tests and flipped out of draft mode. 

AFAICT, this change is not breaking based on current documented behavior (which is pretty minimal). My hope is that this feature is actually short-lived anyway, and replaced with a bigger refactor a la https://github.com/JuliaData/Arrow.jl/issues/88#issuecomment-800744209. 

I thought about adding documentation to accompany this, but then realized that we might not actually want to formally surface it in the API for other folks to start truly relying on - would just cause more churn later when we refactor the underlying system here.

So I've kept this at the same level as some of the other existing machinery here - an undocumented-but-tested "feature" that we can keep at "use-at-your-own-risk" status for now, so that we can replace it more freely later.

jrevels:
superseded by #156 🎉

PR #152, opened 2021-03-14
revert setting Arrow.write debug message threshold to -1

jrevels:
Was surprised to see `DEBUG: ...` messages on `main` even though I had debug level set to `0`. Found the culprit here and assumed checking it in was unintentional; if it WAS intentional feel free to close this PR.


quinnj (approved):
Whoops! Thanks!

Issue #135, opened 2021-02-23
Avoid assuming field values can be used in constructors


omus:
```julia
julia> using Arrow, Intervals

julia> table = (col = [
           Interval{Closed,Unbounded}(1,nothing),
       ],)
(col = Interval{Int64,Closed,Unbounded}[Interval{Int64,Closed,Unbounded}(1, nothing)],)

julia> table.col
1-element Array{Interval{Int64,Closed,Unbounded},1}:
 Interval{Int64,Closed,Unbounded}(1, nothing)

julia> Arrow.write("un.arrow", table);

julia> Arrow.Table("un.arrow").col
1-element Arrow.Struct{Interval{Int64,Closed,Unbounded},Tuple{Arrow.Primitive{Int64,Array{Int64,1}},Arrow.Primitive{Int64,Array{Int64,1}}}}:
Error showing value of type Arrow.Struct{Interval{Int64,Closed,Unbounded},Tuple{Arrow.Primitive{Int64,Array{Int64,1}},Arrow.Primitive{Int64,Array{Int64,1}}}}:
ERROR: MethodError: no method matching Interval{Int64,Closed,Unbounded}(::Int64, ::Int64)
Closest candidates are:
  Interval{Int64,Closed,Unbounded}(::T, ::T, ::Inclusivity) where {T, L, R} at /Users/omus/.julia/dev/Intervals/src/deprecated.jl:34
  Interval{Int64,Closed,Unbounded}(::Any, ::Any, ::Inclusivity) where {T, L, R} at /Users/omus/.julia/dev/Intervals/src/deprecated.jl:44
  Interval{Int64,Closed,Unbounded}(::T, ::Nothing) where {T, L<:Intervals.Bounded, R<:Unbounded} at /Users/omus/.julia/dev/Intervals/src/interval.jl:87
  ...
```
The core of the problem here is that `Interval{T,L,R}` contains two fields `:first` and `:last` which must both be of type `T`. When constructing an unbounded interval we use `nothing` to represent unbounded in the constructor but internally just store an arbitrary value of `T` for the unbounded endpoint. I'll note that native Julia serialization doesn't encounter this problem.

One workaround would be to have Arrow.jl provided callback when constructing the Julia type. That way we could revise the constructor arguments based upon the parametric types.

quinnj:
Ok, with the functionality in #156, we can now support this like:
```julia
using Arrow, Intervals
table = (col = [
[...]

Issue #134, opened 2021-02-23
Loss of parametric type information for custom types


omus:
When using a table where a column contains a variety of types with different parameters this information can be lost:
```julia
julia> using Arrow, Intervals

julia> table = (col = [
           Interval{Closed,Closed}(1,2),
           Interval{Open,Closed}(1,2),
           Interval{Closed,Unbounded}(1,nothing),
       ],);

julia> table.col
3-element Array{Interval{Int64,L,R} where R<:Bound where L<:Bound,1}:
 Interval{Int64,Closed,Closed}(1, 2)
 Interval{Int64,Open,Closed}(1, 2)
 Interval{Int64,Closed,Unbounded}(1, nothing)

julia> Arrow.write("ex.arrow", table);

julia> Arrow.Table("ex.arrow").col
3-element Arrow.Struct{Interval{Int64,L,R} where R<:Bound where L<:Bound,Tuple{Arrow.Primitive{Int64,Array{Int64,1}},Arrow.Primitive{Int64,Array{Int64,1}}}}:
 Interval{Int64,Closed,Closed}(1, 2)
 Interval{Int64,Closed,Closed}(1, 2)
 Interval{Int64,Closed,Closed}(1, 1)
```
For the particular `Interval` type the problem is worse as the undefined type parameters are inferred from the arguments:
```julia
julia> (Interval{Int64,L,R} where R<:Bound where L<:Bound)(1,2)
Interval{Int64,Closed,Closed}(1, 2)
```

quinnj:
Ok, this will now correctly error on #156 PR.

omus:
Using `ArrowTypes.arrowmetadata` as shown in the [Intervals example in the documentation](https://arrow.juliadata.org/dev/manual/#Arrow-types) you can only serialize a column where all of the type parameters are the same. Having mixture of type parameters does not work:
```julia
table = (col = [
    Interval{Closed,Unbounded}(1,nothing),
    Interval{Unbounded,Closed}(nothing,2),
],)
```

omus:
I attempted to work around this on Arrow 1.6 (not quite yet released) by storing the parametric information as part of the value as using `ArrowTypes.arrowmetadata` can't handle element variation. An implementation of this looks like:

```julia
using Arrow, ArrowTypes, Intervals

table = (;
    col=[
        Interval{Closed,Unbounded}(1,nothing),
        Interval{Unbounded,Closed}(nothing,2),
    ]
)

[...]

Issue #132, opened 2021-02-20
`unsupported ARROW:extension:name type: "JuliaLang.Nothing"`


ericphanson:
I ran into a weird situation where I saved an arrow table consisting of a single column of one custom type on one computer (WSL 2 on windows if it matters), then deserialized on another (M1 macbook with Julia running on rosetta), registering the type each time, and I got a bunch of warnings
```julia
┌ Warning: unsupported ARROW:extension:name type: "JuliaLang.Nothing"
└ @ Arrow.ArrowTypes ~/.julia/packages/Arrow/Re9EM/src/arrowtypes.jl:174
```
followed by a related error (`ERROR: MethodError: Cannot `convert` an object of type NamedTuple{(), Tuple{}} to an object of type Symbol`).

I believe I used Arrow 1.2.4 in each case. Manually defining
```julia
julia> Arrow.ArrowTypes.JULIA_TO_ARROW_TYPE_MAPPING[Nothing] = ("JuliaLang.Nothing", Nothing)
("JuliaLang.Nothing", Nothing)

julia> Arrow.ArrowTypes.ARROW_TO_JULIA_TYPE_MAPPING["JuliaLang.Nothing"] = (Nothing, Nothing)
(Nothing, Nothing)
```
then led to the deserialization succeeding (without warnings or errors, and the table seems correct).

When I try serializing and deserializing the exact same type with the exact same setup locally to produce a MWE, I don't get this error. The actual table is here: <https://www.icloud.com/iclouddrive/0fFusdT5pfUior3qaXJ5u3uxA#all_pkgs_results> (3 MB). The code to load it is the following:

```julia
using Pkg
pkg"add https://github.com/giordano/AnalyzeRegistry.jl#990686aaaafc0449151f1adb340cba4c29fe1788" # current master
using Arrow, AnalyzeRegistry
Arrow.ArrowTypes.registertype!(AnalyzeRegistry.Package, AnalyzeRegistry.Package)
load(path) = copy(Arrow.Table(path).packages)
results = load("all_pkgs_results.arrow")
```

Here's the output I get:
```julia
julia> 
┌ Warning: unsupported ARROW:extension:name type: "JuliaLang.Nothing"
└ @ Arrow.ArrowTypes ~/.julia/packages/Arrow/Re9EM/src/arrowtypes.jl:174
┌ Warning: unsupported ARROW:extension:name type: "JuliaLang.Nothing"
[...]

Issue #85, opened 2020-12-23
support for heterogeneously typed tuples


simeonschaub:
Currently writing heterogeneously typed tuples errors:
```julia
julia> Arrow.write("foo.arrow", DataFrame(a = [(1, 2.)]))
ERROR: MethodError: no method matching String(::Int64)
Closest candidates are:
  String(::String) at boot.jl:321
  String(::Array{UInt8,1}) at strings/string.jl:39
  String(::Base.CodeUnits{UInt8,String}) at strings/string.jl:77
  ...
Stacktrace:
 [1] fieldoffset(::Arrow.FlatBuffers.Builder, ::Int64, ::Arrow.Primitive{Int64,Arrow.ToStruct{Int64,1,Array{Tuple{Int64,Float64},1}}}) at /local/scratch/ssd/sschaub/.julia/dev/Arrow/src/write.jl:313
 [2] (::Arrow.var"#74#75"{Arrow.FlatBuffers.Builder,Arrow.Struct{Tuple{Int64,Float64},Tuple{Arrow.Primitive{Int64,Arrow.ToStruct{Int64,1,Array{Tuple{Int64,Float64},1}}},Arrow.Primitive{Float64,Arrow.ToStruct{Float64,2,Array{Tuple{Int64,Float64},1}}}}},Tuple{Int64,Int64}})(::Int64) at ./tuple.jl:0
 [3] iterate at ./generator.jl:47 [inlined]
 [4] collect at ./array.jl:686 [inlined]
 [5] arrowtype(::Arrow.FlatBuffers.Builder, ::Arrow.Struct{Tuple{Int64,Float64},Tuple{Arrow.Primitive{Int64,Arrow.ToStruct{Int64,1,Array{Tuple{Int64,Float64},1}}},Arrow.Primitive{Float64,Arrow.ToStruct{Float64,2,Array{Tuple{Int64,Float64},1}}}}}) at /local/scratch/ssd/sschaub/.julia/dev/Arrow/src/eltypes.jl:389
[...]

PR #156, opened 2021-03-24
Overhaul type serialization/deserialization machinery

quinnj:
Ok, this PR is in response to issues like #135, #134, #132, #88, #85, and a few conversations on slack. In short, the automatic registering of `struct` types is pretty smelly, we're not set up to handle parametric cases very well, and the usage of Dicts internally is fairly un-Julian. We were also pretty inconsistent internally around how things were handled; some types were baked into the `registertype!` machinery (Symbol, Char), some had custom `ArrowType` definitions (UUID), and some were baked in even further to the array conversion code (all the time types). The proposal in this PR is roughly as follows, though I encourage anyone interested to read through the documented interface methods in the arrowtypes.jl file:

* Rename `ArrowType` to `ArrowKind`; it was getting all too muddled together that there is a specific set of native ___types___ that arrow support, and those types all fall under a smaller set of ___physical layout configurations___; the former is now tied to a "new" `ArrowType`, while the latter is now `ArrowKind`. Most custom types don't need to worry about overloading `ArrowKind`, because we define them generally on abstract types in the `ArrowTypes` module already (like `AbstractArray`, `AbstractDict`, etc.). Custom types will more often overload `ArrowType`, which maps a user's custom type to a natively supported arrow type; users should note that custom structs defined like `struct` or `mutable struct`, however, ___are___ natively supported for serialization, but without additional definitions (`arrowname`, `JuliaType`), there is no automatic deserialization
* If I define a custom `ArrowType` definition for my type, then I'm required to also define an `ArrowTypes.toarrow(x::T)` method that converts my type to the native arrow type I defined in `ArrowType`; this provides a serialization "hook" to do any desired transformation; note however, that defining `ArrowTypes.toarrow` ___doesn't___ require having an `ArrowType` definition; I may want to simply ignore a field or two in my custom struct, which I can do by defining `toarrow` and dropping the fields by returning a pared down struct or NamedTuple
* If I want my custom type to be deserializable, I need to implement two methods: `ArrowTypes.arrowname(T) => Symbol`, and `ArrowTypes.JuliaType(::Val{Symbol(name)}, S) = T`. The first takes my custom type as sole argument, and returns a `Symbol` of what my custom type will have stored in the column schema metadata. The latter definition will overload a `Val` wrapping my symbol name from `arrowname` (a not-uncommon Julia technique for value dispatch), and the native arrow serialized type as 2nd argument (`S`). These two definitions allow the "tagging" of my custom type during serialization (`arrowname`) and the conversion from native arrow type to my custom type during deserialization (`JuliaType`). In `JuliaType`, having the native arrow type as 2nd argument allows for parametric types to define a single `arrowname`, and rely on the serialized type to re-parameterize their custom type. Custom types may also choose to just include parameters in their `arrowname` definition; as long as a corresponding `JuliaType` definition exists that exactly matches `Val` argument, all should be well
* The final new interface method is `ArrowTypes.fromarrow`, which provides a deserialization "hook". It uses the result of `JuliaType` definition to provide the native arrow objects as arguments and custom types can then "construct themselves" appropriately.

All in all, I like the power and flexibility of the new system. I think it follows more traditional Julia dispatch patterns, while providing enough customizability to cover necessary use-cases. 

The current test suite passes for me locally, but I have run into an occasional race-condition when reading I believe. I'm going to look more into that. I purposely changed as little as possible in the test suite because I wanted this PR to be as non-breaking as possible. I don't think I'm aware of anyone who was using/defining things with `ArrowTypes.ArrowType` or other internals, which weren't really documented anyway, because I kind of knew it would need an overhaul at some point. I've kept the registertype!` machinery in place for now to avoid breaking things too much.

I'd appreciate any feedback/concerns if people have them. My plan as of now is to go through all the type-related issues and ensure we can now handle the requests, fix any bugs, and add additional tests around some of the new functionality.

### DIFF
--- a/src/arrowtypes.jl
+++ b/src/arrowtypes.jl
@@ -140,6 +140,13 @@ fromarrow(::Type{Union{Missing, T}}, x) where {T} = fromarrow(T, x)
 struct NullKind <: ArrowKind end
 
 ArrowKind(::Type{Missing}) = NullKind()
+ArrowKind(::Type{Nothing}) = NullKind()
+ArrowType(::Type{Nothing}) = Missing
+toarrow(::Nothing) = missing
+const NOTHING = Symbol("JuliaLang.Nothing")
+arrowname(::Type{Nothing}) = NOTHING
+JuliaType(::Val{NOTHING}, S) = Nothing
+fromarrow(::Type{Nothing}, ::Missing) = nothing
 
 "PrimitiveKind data is stored as plain bits in a single contiguous buffer"
 struct PrimitiveKind <: ArrowKind end
@@ -178,6 +185,7 @@ _symbol(ptr, len) = ccall(:jl_symbol_n, Ref{Symbol}, (Ptr{UInt8}, Int), ptr, len
 fromarrow(::Type{Symbol}, ptr::Ptr{UInt8}, len::Int) = _symbol(ptr, len)
 
 ArrowKind(::Type{<:AbstractArray}) = ListKind()
+fromarrow(::Type{A}, x::AbstractVector{T}) where {A <: AbstractVector{T}} where {T} = x
 ArrowKind(::Type{<:AbstractSet}) = ListKind()
 ArrowType(::Type{T}) where {T <: AbstractSet{S}} where {S} = Vector{S}
 toarrow(x::AbstractSet) = collect(x)
--- a/src/arrowtypes.jl
+++ b/src/arrowtypes.jl
@@ -22,75 +22,170 @@ module ArrowTypes
 
 using UUIDs
 
-export ArrowKind, NullType, PrimitiveType, BoolType, ListType, FixedSizeListType, MapType, StructType, UnionType, DictEncodedType, toarrow, arrowname, fromarrow, ToArrow
+export ArrowKind, NullKind, PrimitiveKind, BoolKind, ListKind, FixedSizeListKind, MapKind, StructKind, UnionKind, DictEncodedKind, toarrow, arrowname, fromarrow, ToArrow
 
+"""
+    ArrowTypes.ArrowKind(T)
+
+For a give type `T`, define it's "arrow type kind", or the general category of arrow types it should be treated as. Must be one of:
+  * [`ArrowTypes.NullKind`](@ref): `Missing` is the only type defined as `NullKind`
+  * [`ArrowTypes.PrimitiveKind`](@ref): `<:Integer`, `<:AbstractFloat`, along with `Arrow.Decimal`, and the various `Arrow.ArrowTimeType` subtypes
+  * [`ArrowTypes.BoolKind`](@ref): only `Bool`
+  * [`ArrowTypes.ListKind`](@ref): any `AbstractString` or `AbstractArray`
+  * [`ArrowTypes.FixedSizeList`](@ref): `NTuple{N, T}`
+  * [`ArrowTypes.MapKind`](@ref): any `AbstractDict`
+  * [`ArrowTypes.StructKind`](@ref): any `NamedTuple` or plain struct (mutable or otherwise)
+  * [`ArrowTypes.UnionKind`](@ref): any `Union`
+  * [`ArrowTypes.DictEncodedKind`](@ref): array types that implement the `DataAPI.refarray` interface
+
+The list of `ArrowKind`s listed above translate to different ways to physically store data as supported by the arrow data format.
+See the docs for each for an idea of whether they might be an appropriate fit for a custom type.
+Note that custom types need to satisfy any additional "interface methods" as required by the various `ArrowKind`
+types. By default, if a type in julia is declared like `primitive type ...` it is considered a `PrimitiveKind`
+and if `struct` or `mutable struct` it's considered a `StructKind`. Also note that types will rarely need to define `ArrowKind`;
+much more common is to define `ArrowType(T)` and `toarrow(x::T)` to transform `T` to a natively supported arrow type, which will
+already have its `ArrowKind` defined.
+"""
 abstract type ArrowKind end
 
 ArrowKind(x::T) where {T} = ArrowKind(T)
-ArrowKind(::Type{T}) where {T} = isprimitivetype(T) ? PrimitiveType() : StructType()
+ArrowKind(::Type{T}) where {T} = isprimitivetype(T) ? PrimitiveKind() : StructKind()
 
+"""
+    ArrowTypes.ArrowType(T) = S
+
+Interface method to define the natively supported arrow type `S` that a given type `T` should be converted to before serializing.
+Useful when a custom type wants a "serialization hook" or otherwise needs to be transformed/converted into a natively
+supported arrow type for serialization. If a type defines `ArrowType`, it must also define a corresponding
+[`ArrowTypes.toarrow(::Type{S}, x::T)`](@ref) method which does the actual conversion from `T` to `S`.
+Note that custom structs defined like `struct T` or `mutable struct T` are natively supported in serialization, so unless
+_additional_ transformation/customization is desired, a custom type `T` can serialize with no `ArrowType` definition (by default,
+each field of a struct is serialized, using the results of `fieldnames(T)` and `getfield(x, i)`).
+Note that defining these methods only enables _serialization_ to the arrow format; to be able to _deserialize_ custom types,
+see the docs for [`ArrowTypes.arrowname`](@ref), [`ArrowTypes.JuliaType`](@ref), and [`ArrowTypes.fromarrow`](@ref).
+"""
 function ArrowType end
 ArrowType(::Type{T}) where {T} = T
 
-function toarrow end
-toarrow(T, x) = x
+"""
+    ArrowTypes.toarrow(x::T) => S
 
+Interface method to perform the actual conversion from an object `x` of type `T` to the type `S`. `T` and `S` must match the
+types used when defining `ArrowTypes.ArrowType(::Type{T}) = S`. See [`ArrowTypes.ArrowType`](@ref) docs for more details.
+This enables custom objects to be serialized as a natively supported arrow data type.
+"""
+function toarrow end
+toarrow(x) = x
+
+"""
+    ArrowTypes.arrowname(T) = Symbol(name)
+
+Interface method to define the logical type "label" for a custom Julia type `T`. Names will be global for an entire arrow dataset,
+and conventionally, custom types will just use their type name along with a Julia-specific prefix; for example, for a custom type
+`Foo`, I would define `ArrowTypes.arrowname(::Type{Foo}) = Symbol("JuliaLang.Foo")`. This ensures other language implementations
+won't get confused and are safe to ignore the logical type label.
+When arrow stores non-native data, it must still be _stored_ as a native data type, but can have metadata tied to the data that
+labels the original _logical_ type it originated from. This enables the conversion of native data back to the logical type when
+deserializing, as long as the deserializer has the same definitions when the data was serialized. Namely, the current Julia
+session will need the appropriate [`ArrowTypes.JuliaType`](@ref) and [`ArrowTypes.fromarrow`](@ref) definitions in order to know
+how to convert the native data to the original logical type. See the docs for those interface methods in order to ensure a complete implementation.
+"""
 function arrowname end
 const EMPTY_SYMBOL = Symbol()
 arrowname(T) = EMPTY_SYMBOL
+hasarrowname(T) = arrowname(T) !== EMPTY_SYMBOL
 
+"""
+    ArrowTypes.JuliaType(::Val{Symbol(name)}, ::Type{S}) = T
+
+Interface method to define the custom Julia logical type `T` that a serialized metadata label should be converted to when
+deserializing. When reading arrow data, it may encounter a logical type label for a column, it will call
+`ArrowTypes.JuliaType(Val(Symbol(name)), S)` to see if a Julia type has been "registered" for deserialization. The `name`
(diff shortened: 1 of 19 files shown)


jrevels:
if I get a chance in the next few days I'll try this out over in https://github.com/beacon-biosignals/Onda.jl/pull/68 (my initial use case) and report back

src/arrowtypes.jl
```diff
@@ -22,56 +22,180 @@ module ArrowTypes
...
 
-ArrowType(::Type{<:AbstractString}) = ListType()
-isstringtype(::Type{<:AbstractString}) = true
+"""
+    ArrowTypes.ArrowType(T) = S
+
+Interface method to define the natively supported arrow type `S` that a given type `T` should be converted to before serializing.
+Useful when a custom type wants a "serialization hook" or otherwise needs to be transformed/converted into a natively
+supported arrow type for serialization. If a type defines `ArrowType`, it must also define a corresponding
+[`ArrowTypes.toarrow(::Type{S}, x::T)`](@ref) method which does the actual conversion from `T` to `S`.
```
jrevels, line 60:
```suggestion
[`ArrowTypes.toarrow(x::T)`](@ref) method which does the actual conversion from `T` to `S`.
```

src/arrowtypes.jl
```diff
@@ -22,56 +22,180 @@ module ArrowTypes
...
+
+Interface method to define the custom Julia logical type `T` that a serialized metadata label should be converted to when
+deserializing. When reading arrow data, it may encounter a logical type label for a column, it will call
+`ArrowTypes.JuliaType(Val(Symbol(name)), S)` to see if a Julia type has been "registered" for deserialization. The `name`
+used when defining the method *must* correspond to the same `name` when defining `ArrowTypes.arrowname(::Type{T}) = Symbol(name)`.
+The use of `Val(Symbol(...))` is to allow overloading a method on a specific logical type label. The `S` 2nd argument passed to
+`JuliaType` is the native arrow serialized type. This can be useful for parametric Julia types that wish to correctly parameterize
+their custom type based on what was serialized.
+When defining [`ArrowTypes.arrowname`](@ref) and `ArrowTypes.JuliaType`, you may also want to implement [`ArrowTypes.fromarrow`]
+in order to customize how a custom type `T` should be constructed from the native arrow data type. See its docs for mmore details.
```
jrevels, line 111:
```suggestion
in order to customize how a custom type `T` should be constructed from the native arrow data type. See its docs for more details.
```

src/arrowtypes.jl
```diff
@@ -22,56 +22,180 @@ module ArrowTypes
...
+Interface method that provides a "deserialization hook" for a custom type `T` to be constructed from the native arrow type `S`.
+The `T` and `S` types must correspond to the definitions used in `ArrowTypes.ArrowType(::Type{T}) = S`. This is a paired method
+with [`ArrowTypes.toarrow`](@ref).
+
+The default definition is `ArrowTypes.fromarrow(::Type{T}, x) = T(x)`, so if that works for a custom type already, no additional
+overload is necessary.
+A few `ArrowKind`s allow slightly more custom overloads for their `fromarrow` methods:
+  * `ListKind{true}`: for `String` types, they may overload `fromarrow(::Type{T}, ptr::Ptr{UInt8}, len::Int) = ...` to avoid
+     materializing a `String`
+  * `StructKind`: may overload `fromarrow(::Type{T}, x::NamedTuple)`, or `fromarrow(::Type{T}; kw...)` (values passed as
```
jrevels, line 128: over in https://github.com/beacon-biosignals/Onda.jl/pull/68/files#diff-4c48d3b8e6e0b82d379c756b160263e0a23aabbb82a4f1fadfbbf02f35b0814cR304

I have a `fromarrow(::Type{T}, x::NamedTuple)` method but it doesn't seem to be getting called:

```
julia> include("examples/tour.jl"); # run from root of Onda directory, generates some test state

julia> infos = [eeg.info, eeg.info, eeg.info, eeg.info]; # dummy data

julia> using Arrow

julia> tbl = Arrow.Table(Arrow.tobuffer((x=infos,)))
Arrow.Table: (x = SamplesInfo{String,Array{String,1},String,Float64,Int64,var"#s89",Float64} where var"#s89"<:Union{Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8}[Error showing value of type Arrow.Table:
ERROR: MethodError: no method matching SamplesInfo{String,Array{String,1},String,Float64,Int64,var"#s89",Float64} where var"#s89"<:Union{Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8}(::String, ::Array{String,1}, ::String, ::Float64, ::Int64, ::String, ::Float64)
Stacktrace:
 [1] fromarrow(::Type{SamplesInfo{String,Array{String,1},String,Float64,Int64,var"#s89",Float64} where var"#s89"<:Union{Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8}}, ::String, ::Array{String,1}, ::String, ::Vararg{Any,N} where N) at /Users/jarrettrevels/.julia/dev/Arrow/src/arrowtypes.jl:134
 [2] getindex at /Users/jarrettrevels/.julia/dev/Arrow/src/arraytypes/struct.jl:47 [inlined]
 [3] isassigned(::Arrow.Struct{SamplesInfo{String,Array{String,1},String,Float64,Int64,var"#s89",Float64} where var"#s89"<:Union{Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8},Tuple{Arrow.List{String,Int32,Array{UInt8,1}},Arrow.List{Array{String,1},Int32,Arrow.List{String,Int32,Array{UInt8,1}}},Arrow.List{String,Int32,Array{UInt8,1}},Arrow.Primitive{Float64,Array{Float64,1}},Arrow.Primitive{Int64,Array{Int64,1}},Arrow.List{String,Int32,Array{UInt8,1}},Arrow.Primitive{Float64,Array{Float64,1}}}}, ::Int64) at ./abstractarray.jl:408
 [4] show_delim_array(::IOContext{REPL.Terminals.TTYTerminal}, ::Arrow.Struct{SamplesInfo{String,Array{String,1},String,Float64,Int64,var"#s89",Float64} where var"#s89"<:Union{Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8},Tuple{Arrow.List{String,Int32,Array{UInt8,1}},Arrow.List{Array{String,1},Int32,Arrow.List{String,Int32,Array{UInt8,1}}},Arrow.List{String,Int32,Array{UInt8,1}},Arrow.Primitive{Float64,Array{Float64,1}},Arrow.Primitive{Int64,Array{Int64,1}},Arrow.List{String,Int32,Array{UInt8,1}},Arrow.Primitive{Float64,Array{Float64,1}}}}, ::Char, ::String, ::Char, ::Bool, ::Int64, ::Int64) at ./show.jl:740
 [5] show_delim_array at ./show.jl:733 [inlined]
 [6] show_vector(::IOContext{REPL.Terminals.TTYTerminal}, ::Arrow.Struct{SamplesInfo{String,Array{String,1},String,Float64,Int64,var"#s89",Float64} where var"#s89"<:Union{Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8},Tuple{Arrow.List{String,Int32,Array{UInt8,1}},Arrow.List{Array{String,1},Int32,Arrow.List{String,Int32,Array{UInt8,1}}},Arrow.List{String,Int32,Array{UInt8,1}},Arrow.Primitive{Float64,Array{Float64,1}},Arrow.Primitive{Int64,Array{Int64,1}},Arrow.List{String,Int32,Array{UInt8,1}},Arrow.Primitive{Float64,Array{Float64,1}}}}, ::Char, ::Char) at ./arrayshow.jl:476
 [7] show_vector at ./arrayshow.jl:461 [inlined]
 [8] show at ./arrayshow.jl:432 [inlined]
 [9] show(::IOContext{REPL.Terminals.TTYTerminal}, ::NamedTuple{(:x,),Tuple{Arrow.Struct{SamplesInfo{String,Array{String,1},String,Float64,Int64,var"#s89",Float64} where var"#s89"<:Union{Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8},Tuple{Arrow.List{String,Int32,Array{UInt8,1}},Arrow.List{Array{String,1},Int32,Arrow.List{String,Int32,Array{UInt8,1}}},Arrow.List{String,Int32,Array{UInt8,1}},Arrow.Primitive{Float64,Array{Float64,1}},Arrow.Primitive{Int64,Array{Int64,1}},Arrow.List{String,Int32,Array{UInt8,1}},Arrow.Primitive{Float64,Array{Float64,1}}}}}}) at ./namedtuple.jl:150
 [10] show(::IOContext{REPL.Terminals.TTYTerminal}, ::Arrow.Table) at /Users/jarrettrevels/.julia/dev/Tables/src/Tables.jl:196
 [11] show(::IOContext{REPL.Terminals.TTYTerminal}, ::MIME{Symbol("text/plain")}, ::Arrow.Table) at ./multimedia.jl:47
 [12] display(::REPL.REPLDisplay, ::MIME{Symbol("text/plain")}, ::Any) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:214
 [13] display(::REPL.REPLDisplay, ::Any) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:218
 [14] display(::Any) at ./multimedia.jl:328
 [15] #invokelatest#1 at ./essentials.jl:710 [inlined]
 [16] invokelatest at ./essentials.jl:709 [inlined]
 [17] print_response(::IO, ::Any, ::Bool, ::Bool, ::Any) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:238
 [18] print_response(::REPL.AbstractREPL, ::Any, ::Bool, ::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:223
 [19] (::REPL.var"#do_respond#54"{Bool,Bool,REPL.var"#64#73"{REPL.LineEditREPL,REPL.REPLHistoryProvider},REPL.LineEditREPL,REPL.LineEdit.Prompt})(::Any, ::Any, ::Any) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:822
 [20] #invokelatest#1 at ./essentials.jl:710 [inlined]
 [21] invokelatest at ./essentials.jl:709 [inlined]
 [22] run_interface(::REPL.Terminals.TextTerminal, ::REPL.LineEdit.ModalInterface, ::REPL.LineEdit.MIState) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/LineEdit.jl:2355
 [23] run_frontend(::REPL.LineEditREPL, ::REPL.REPLBackendRef) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:1144
 [24] (::REPL.var"#38#42"{REPL.LineEditREPL,REPL.REPLBackendRef})() at ./task.jl:356
```

Looks like it's splatting the arguments instead?

quinnj: Yeah, I realized it would be quite a bit more inefficient to generate the full NamedTuple. My hope was that the compiler would optimize it out entirely, but I would need to comb over the code to see if that's even possible. We'd probably have to include the arrow type as another type parameter to the Struct array type. That might be a good idea anyways since I'm currently deep in the weeds on a big where it gets pretty confusing if the Struct eltype is Julia or arrow.

quinnj:
Still to do:
* Still working on supporting https://github.com/JuliaData/Arrow.jl/issues/132; it's a doozy and that's really good because it's really stress testing all the type serialization stuff; there's some kind of bug either writing or reading the `lines_of_code` and `contributors` fields that's causing things to get messed up. It's been tricky to minimally reproduce, so I'll have to spend some more time on it tomorrow
* Another pass over docs; things have evolved a bit and I still need to update the proper manual on all the new custom types stuff
* More tests; I'm still working through open issues making sure we fix/support everything, but I'd like to do some additional unit tests on top of that as well

quinnj:
Ok, just pushed a new commit that allows us to properly support cases like https://github.com/JuliaData/Arrow.jl/issues/135. The wrinkle there is the custom type `Interval` has non-field-inferred/based type parameters. You could potentially support that by just having separate `ArrowTypes.arrowname`/`ArrowTypes.JuliaType` definitions for different type parameter combinations. That's probably reasonable if the # of unique type parameters is low, but otherwise, becomes impractical pretty quickly.

In the [arrow spec about extension types](https://arrow.apache.org/docs/format/Columnar.html#extensibility), it also officially supports setting the `"ARROW:extension:metadata"` key with a string value to provide additional extension type metadata needed for recreating the original logical type. That seems to fit pretty well with our need, so I've added an optional `ArrowTypes.arrowmetadata(T)` method users can overload to provide any additional type information necessary when deserializing. This is then available by overloading `ArrowTypes.JuliaType(::Val{Symbol(name)}, S, arrowmetadata)` and using `arrowmetadata` however necessary to recreate the types. I gave an example of this for Intervals [here](https://github.com/JuliaData/Arrow.jl/issues/135#issuecomment-807252002).

quinnj:
Ok, I've updated some docs + the manual and added some more tests. I think this is ready to go. I'll leave it open for another day or two for others to try things out if they want, but then I think we're ready for a merge + 1.3 release.

Issue #105, opened 2021-01-10
Allow for file appends


ayman-albaz:
Hello,

I'm currently dealing with a problem where I need to be able to append to a *.arrow file. The data needs to be added row by row because its bigger than memory and can only be loaded row by row through a stream.

A good example of what I want to do is this.

```julia
using Arrow
using DataFrames
using Tables

open("out.arrow", "a") do file
    for i in massive_stream_iterator
        X = rand(1, 1000) |> Tables.table
        Arrow.write(file, X)
    end
end
```

Reading from this file generates a 1x1000 sized table not an Nx1000 sized table. 
```julia
df = Arrow.Table("append.arrow") |> DataFrame
@info size(df)
```

Is it currently possible for Arrow.jl to do this? I've looked all over the documentation and the closest thing to this was using Table.partitioner, which I can't use since I'm iterating through a massive-bigger-than-memory iterator.

quinnj:
So the tricky thing here is that the arrow format is _columnar_; so rows have to be collected and columns have to be "built" at some point to write the data out. i.e. it's not physically possible to write out an arrow file row-by-row.

Now, we do have some new tools in the ecosystem that might help in this case. For arrow, if you're working with larger-than-memory data, the solution is to write the data in "batches", which `Arrow.write` supports when the table source implements `Tables.partitions`. This allows writing out each "partition" as a separate record batch in the arrow file.

So you're correct in thinking `Tables.partitioner` could help here, it allows "partitioning" an input lazily, like a lazy `map` that produces a valid table on each iteration. So the usage in your example above would be more like:
```julia
Arrow.write("out.arrow", Tables.partitioner(x->Iterators.take(x, 1000), massive_stream_iterator) )
```
What this will do is spin up a writing thread for each of `Threads.nthreads()`, each will call `Iterators.take(massive_stream_iterator, 1000)`, which will iterate/return 1000 elements, those 1000 elements will be made into columns and written out as a single record batch. So the `massive_stream_iterator` will be written out in record batches of 1000 elements each. This assumes that each element iterated from `massive_stream_iterator` is a valid "row" according to the Tables.jl interface (i.e. some object that supports at least `propertynames(x)` and `getproperty(x, nm)`).

Anyway, let me know if that sounds like something in the right direction. I think it could also be possible to support something like `Arrow.append(file, x)` if we really needed: we would open `file` and parse the existing schema and then verify that the new table `x` matches that schema and then write the single table out as a record batch.

[...]

PR #160, opened 2021-04-01
ability to append partitions to existing arrow files

tanmaykm:
This adds a method to `append` partitions to existing arrow files. Partitiions to append to are supplied in the form of any [Tables.jl](https://github.com/JuliaData/Tables.jl)-compatible table.

Multiple record batches will be written based on the number of `Tables.partitions(tbl)` that are provided.

Each partition being appended must have the same `Tables.Schema` as the destination arrow file that is being appended to.

Other parameters that `append` accepts are similar to what `write` accepts.

### DIFF
--- a/src/append.jl
+++ b/src/append.jl
@@ -187,14 +187,6 @@ function table_info(bytes::Vector{UInt8}; convert::Bool=true)
     Tables.Schema(names, types), dictencodings, compression_codec
 end
 
-function footer_offset(bytes::Vector{UInt8})
-    magic_len = length(FILE_FORMAT_MAGIC_BYTES)
-    startoffset = length(bytes) - magic_len - sizeof(Int32) + 1
-    endoffset = length(bytes) - magic_len
-
-    read(IOBuffer(bytes[startoffset:endoffset]), Int32) + magic_len + sizeof(Int32)
-end
-
 function is_stream_format(file::String)
     open(file) do io
         is_stream_format(io, true)
@@ -211,31 +203,6 @@ function is_stream_format(bytes::Vector{UInt8})
     else
         return true
     end
-end
-
-function convert_to_stream_format(file::String; kwargs...)
-    mktemp() do path, io
-        convert_to_stream_format(file, io; kwargs...)
-        close(io)
-        cp(path, file; force=true)
-    end
-    nothing
-end
-function convert_to_stream_format(source::String, dest::String; kwargs...)
-    open(dest, "w") do destio
-        convert_to_stream_format(source, destio; kwargs...)
-    end
-    nothing
-end
-function convert_to_stream_format(source::String, destio::IO; kwargs...)
-    open(source) do sourceio
-        convert_to_stream_format(sourceio, destio; kwargs...)
-    end
-    nothing
-end
-function convert_to_stream_format(sourceio::IO, destio::IO; kwargs...)
-    write(destio, Stream(sourceio); file=false, kwargs...)
-    nothing
 end
 
 function is_equivalent_schema(sch1::Tables.Schema, sch2::Tables.Schema)
--- a/src/append.jl
+++ b/src/append.jl
@@ -105,16 +105,17 @@ function append(io::IO, source, arrow_schema, compress, largelists, denseunions,
             error("fatal error writing arrow data")
         end
         @debug 1 "processing table partition i = $i"
-        tbl_schema = Tables.schema(tbl)
+        tbl_cols = Tables.columns(tbl)
+        tbl_schema = Tables.schema(tbl_cols)
 
         if !is_equivalent_schema(arrow_schema, tbl_schema)
             throw(ArgumentError("Table schema does not match existing arrow file schema"))
         end
 
         if threaded
-            Threads.@spawn process_partition(tbl, dictencodings, largelists, compress, denseunions, dictencode, dictencodenested, maxdepth, msgs, alignment, i, sch, errorref, anyerror)
+            Threads.@spawn process_partition(tbl_cols, dictencodings, largelists, compress, denseunions, dictencode, dictencodenested, maxdepth, msgs, alignment, i, sch, errorref, anyerror)
         else
-            @async process_partition(tbl, dictencodings, largelists, compress, denseunions, dictencode, dictencodenested, maxdepth, msgs, alignment, i, sch, errorref, anyerror)
+            @async process_partition(tbl_cols, dictencodings, largelists, compress, denseunions, dictencode, dictencodenested, maxdepth, msgs, alignment, i, sch, errorref, anyerror)
         end
     end
     if anyerror[]
--- a/src/append.jl
+++ b/src/append.jl
@@ -63,11 +63,10 @@ function append(io::IO, tbl;
         throw(ArgumentError("ntasks keyword argument must be > 0; pass `ntasks=1` to disable multithreaded writing"))
     end
 
-    if !is_stream_format(io)
+    isstream, arrow_schema, compress = stream_properties(io; convert=convert)
+    if !isstream
         throw(ArgumentError("append is supported only to files in arrow stream format"))
     end
-
-    arrow_schema, compress = table_info(io; file=file, convert=convert)
 
     if compress === :lz4
         compress = LZ4_FRAME_COMPRESSOR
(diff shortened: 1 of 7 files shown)


tanmaykm:
Also ref #105 which seems related.

src/append.jl
```diff
@@ -0,0 +1,168 @@
+"""
+    Arrow.append(file::String, tbl)
```
quinnj, line 2:
```suggestion
    Arrow.append(io_or_file, tbl)
```

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+    tbl |> Arrow.append(io_or_file)
+
+Append any [Tables.jl](https://github.com/JuliaData/Tables.jl)-compatible
+`tbl` to an existing arrow formatted file.
```
quinnj, line 6:
```suggestion
`tbl` to an existing arrow formatted IPC stream, as any kind of `IO` or if given a `String`, assume the IPC stream has been written to disk. Note that appending to the "feather formatted file" is _not_ allowed, as this file format doesn't support appending. That means files written like `Arrow.write(filename::String, tbl)` _cannot_ be appended to; instead, you should write like `Arrow.write(filename::String, tbl; file=false)`.
```

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+  * `denseunions::Bool=true`: whether Julia `Vector{<:Union}` arrays should be written using the dense union layout; passing `false` will result in the sparse union layout
+  * `largelists::Bool=false`: causes list column types to be written with Int64 offset arrays; mainly for testing purposes; by default, Int64 offsets will be used only if needed
+  * `maxdepth::Int=$DEFAULT_MAX_DEPTH`: deepest allowed nested serialization level; this is provided by default to prevent accidental infinite recursion with mutually recursive data structures
+  * `ntasks::Int`: number of concurrent threaded tasks to allow while writing input partitions out as arrow record batches; default is no limit; to disable multithreaded writing, pass `ntasks=1`
+"""
+function append end
+
+append(file::String; kw...) = x -> append(file::String, x; kw...)
+
+function append(file::String, tbl;
```
quinnj, line 33: we should copy the `Arrow.write` definitions we have that allow `::IO` as well. We should also have a functor form like `append(io_or_file; kw...) = x -> append(io_or_file, x; kw...)

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+        largelists::Bool=false,
+        compress::Union{Nothing, Symbol, LZ4FrameCompressor, ZstdCompressor}=nothing,
```
quinnj, line 35: We don't want to allow passing in this since we'll need to match the compression used in the rest of the file.

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+        denseunions::Bool=true,
+        dictencode::Bool=false,
+        dictencodenested::Bool=false,
+        alignment::Int=8,
```
quinnj, line 39: Same here; we need to match the alignment used in the rest of the file

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+
+    open(file, "r+") do io
+        append(io, tbl, largelists, compress, denseunions, dictencode, dictencodenested, alignment, maxdepth, ntasks)
+    end
+
+    return file
+end
+
+function append(io, source, largelists, compress, denseunions, dictencode, dictencodenested, alignment, maxdepth, ntasks)
+    bytes = Mmap.mmap(io)
```
quinnj, line 61: Hmmm, I don't think this works if you provide a filename? Because you'd be calling `bytes = Mmap.mmap(io::IOStream)`? maybe that's supported? I'd rather follow the pattern we have in `Arrow.Table` where we call `Mmap.mmap` if they provide a `string`, otherwise we call `read(io)` to get the IPC stream bytes.

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+    stream = Arrow.Stream(bytes)
+    arrow_schema = Tables.schema(Tables.columns(stream))
```
quinnj, line 63: If this works, then it's probably by accident. By design, you're supposed to call `Tables.columns` on each _iterated element_ of an `Arrow.Stream`. We can, however, access the arrow schema by calling `stream.schema`. That gives us the arrow schema as a flatbuffer object. I think that's the right thing we need to validate against. For example, if we start iterating input `Tables.partitions(source)`, then we call `cols = Tables.columns(part); sch = Tables.schema(cols)`, then we can call `makeschemamsg(sch, cols)` which gives us back a flatbuffer of the schema. Then we should flip that flatbuffer around to read it `arrowsch = Meta.Schema(buf, 1)` (hopefully that works). Then we can compare `stream.schema` vs `arrowsch`.

quinnj: The alternative approach would be to iterate the first table of the `Arrow.Stream`, the call `sch = Tables.schema(Tables.columns(first(stream)))` and compare that w/ each partition schema we iterate. Probably a bit simpler than the arrow flatbuffer schema comparing.

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+    end)
+    anyerror = Threads.Atomic{Bool}(false)
+    errorref = Ref{Any}()
+    @sync for (i, tbl) in enumerate(Tables.partitions(source))
+        if anyerror[]
+            @error "error writing arrow data on partition = $(errorref[][3])" exception=(errorref[][1], errorref[][2])
+            error("fatal error writing arrow data")
+        end
+        @debug 1 "processing table partition i = $i"
+        tbl_schema = Tables.schema(tbl)
```
quinnj, line 94: Technically, we need to call `Tables.schema` on the result of `Tables.columns(tbl)`, though often they produce the same result. But we should probably pass the `stream` schema down to `process_partition` to do the comparison there.

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+    if anyerror[]
+        @error "error writing arrow data on partition = $(errorref[][3])" exception=(errorref[][1], errorref[][2])
+        error("fatal error writing arrow data")
+    end
+    # close our message-writing channel, no further put!-ing is allowed
+    close(msgs)
+    # now wait for our message-writing task to finish writing
+    wait(tsk)
+
+    b = FlatBuffers.Builder(1024)
```
quinnj, line 115: we shouldn't need any of this code; it's not going to be valid to append to a feather formatted file, which needs this footer stuff.

src/append.jl
```diff
@@ -0,0 +1,168 @@
...
+    startoffset = length(bytes) - magic_len - sizeof(Int32) + 1
+    endoffset = length(bytes) - magic_len
+
+    read(IOBuffer(bytes[startoffset:endoffset]), Int32) + magic_len + sizeof(Int32)
+end
+
+function is_equivalent_schema(sch1::Tables.Schema, sch2::Tables.Schema)
+    (sch1.names == sch2.names) || (return false)
+    for (t1,t2) in zip(sch1.types, sch2.types)
+        (t1 <: Arrow.ArrowVector) && (t1 = eltype(t1))
```
quinnj, line 163: we shouldn't need this extra step after we fix the above code where we get tehs chema of the first table iterated by the stream

src/append.jl
tanmaykm: How can we retrieve this from the supplied `io_or_file`?

src/append.jl
tanmaykm: Can't we append to a feather/arrow ipc file format by starting to write from the footer offset? Then we need to write the new footer back at the end, no?

src/append.jl
tanmaykm: done in https://github.com/JuliaData/Arrow.jl/pull/160/commits/6a56db3ad923e97a55f7131a86698a230ebcd9bd

src/append.jl
tanmaykm: done in https://github.com/JuliaData/Arrow.jl/pull/160/commits/6a56db3ad923e97a55f7131a86698a230ebcd9bd

src/append.jl
quinnj: Hmmm, that sounds quite a bit more complicated. We'd have to _read_ the existing footer, because we need all that existing info, then we'd need to write out the new footer; and yes, rewrite over the existing one. I think that could be a separate PR, where we first start with only support appending to IPC streams.

src/append.jl
quinnj: Looks like we'll need to read the first record batch message from the `Arrow.Stream`, then we can see the `compression` on the RecordBatch message (property is here: https://github.com/JuliaData/Arrow.jl/blob/main/src/metadata/Message.jl#L76)

src/append.jl
tanmaykm: How can the alignment be retrieved from `io_or_file`? Maybe I am missing it, but I do not see that getting stored explicitly anywhere.

src/append.jl
tanmaykm: Actually, appending to stored files is the functionality I need and that is what I tried to address here. But looks like it is valid to store arrow data into files in stream format? Then we don't need to deal with the footer which will make things simpler. Is that what you meant here?

src/append.jl
tanmaykm: changed in https://github.com/JuliaData/Arrow.jl/pull/160/commits/7aca33da479fad8e98d1ff192d98adf9aba6e8b7 now

src/append.jl
tanmaykm: done in https://github.com/JuliaData/Arrow.jl/pull/160/commits/7aca33da479fad8e98d1ff192d98adf9aba6e8b7

src/append.jl
tanmaykm: done that way in https://github.com/JuliaData/Arrow.jl/pull/160/commits/7aca33da479fad8e98d1ff192d98adf9aba6e8b7 now

quinnj: Yes, that's what I meant. It's much simpler to append to file when data is in IPC format instead of "file" format.

src/append.jl
quinnj: Ok, now that I think about it and look back at code, I think we can leave this as-is. It's actually valid to use different alignment paddings in different record batches, because the buffers are all specified in full byte counts. So the alignment is "implicit", and shouldn't matter.

src/append.jl
tanmaykm: The functor form is now present for `append`. It does not yet accept an IO but can be added with a caveat that I wanted to put up.

It seems that we need to overwrite the last empty message in an arrow stream when we append to it, so if we accept an IO, we need to be able to `seek` on it to be able to go back to the point where we got the empty message. And the API will be writing the record batches it appends on to the same IO stream. We will be able to do this on a stream opened on a file, or an IOBuffer, but we will not be able to do it on a network stream for example. Will that be acceptable?

tanmaykm:
Hi @quinnj, does this look okay, or is anything more needed here?

quinnj:
Sorry, I've been trying to catch up on a bunch of stuff since coming back from vacation; planning on reviewing this more in-depth in the next 24 hours.

src/append.jl
quinnj: Yes, those restrictions seem fine to me and worth mentioning in the docs.

src/append.jl
```diff
@@ -0,0 +1,214 @@
...
+    return io
+end
+
+function table_info(file::String; kwargs...)
+    open(file) do io
+        table_info(io; file=true, kwargs...)
+    end
+end
+table_info(io::IO; file::Bool=false, kwargs...) = table_info(file ? Mmap.mmap(io) : Base.read(io); kwargs...)
+function table_info(bytes::Vector{UInt8}; convert::Bool=true)
```
quinnj, line 120: Hmmmm, I don't like duplicating all this code from `Arrow.Stream`/`Arrow.Table`. I think we should modify those to store whatever is needed: schema, compression, etc. I'd much rather add a field or two to `Arrow.Stream` rather than have all this duplicate code.

tanmaykm: Yes, good idea. Will make the change.

tanmaykm: done in https://github.com/JuliaData/Arrow.jl/pull/160/commits/24d5920cacd845c13d18e3dbf640ba4a64dd552c

src/append.jl
tanmaykm: addressed in https://github.com/JuliaData/Arrow.jl/pull/160/commits/00da06d181c9f2d385964cd450315dcbcb5f455b

src/append.jl
tanmaykm: done in https://github.com/JuliaData/Arrow.jl/pull/160/commits/00da06d181c9f2d385964cd450315dcbcb5f455b

tanmaykm:
@quinnj I have now added commits to allow appends to `IO` and added few members to `Stream` instead of duplicating code for `append`.

src/append.jl
quinnj: Ok, I think this all looks good now; this is the only remaining piece. Specifically, we're calling `Tables.schema(tbl)`, where `tbl` is a partition iteration element from `Tables.partitions`. Technically, we need to call `Tables.schema(Tables.columns(tbl))`. We don't end up calling `Tables.columns` until `process_partition`. Probably easiest is to change `process_partition` to expect the 1st argument to already be the result of `Tables.columns(tbl)` and we can call it like that here:
```julia
cols = Tables.columns(tbl)
tbl_schema = Tables.schema(cols)
...
process_partition(cols, ...)
```
Then we just need to make sure we update the usage of `process_partition` in write.jl as well.

tanmaykm: done in https://github.com/JuliaData/Arrow.jl/pull/160/commits/0f0ab4bbfc2f72dc7e1ac41071a84465f7c58c01

tanmaykm:
bump!

quinnj (approved):
Thanks for all the work on this @tanmaykm!

PR #170, opened 2021-04-07
add missing setmedata! method for Arrow.Table

jrevels:
I'm not sure if I'm misinterpreting, but it seems like #165 also needed a corresponding `setmetadata!` for the specialization to behave as expected - otherwise the test added by this PR fails (and it seems like such a test should pass 😁)


quinnj (approved):
Yes, thanks for this!

PR #171, opened 2021-04-07
use actual deprecation for `registertype!`

ericphanson:
This means if you aren't running tests or otherwise start julia with `--depwarn=yes` then you don't see the warning, and moreover you only see the warning once. Motivated by me seeing this warning three times when just calling `using X` from some downstream package that doesn't even call `registertype!` itself!

If you do want it to be visible to all users to try to push more for users to update, then another option is to use `@warn` but with `maxlog=1` which should emit the warning only once.


quinnj (approved):
Thanks; I can never quite remember the exact right incantation here.

PR #174, opened 2021-04-14
Fix () -> {} typo in docs

etpinard:
to avoid

```jl
julia> ArrowTypes.JuliaType(::Val(:Person)) = Person
ERROR: ArgumentError: invalid type for argument number 1 in method definition for JuliaType at REPL[13]:1
Stacktrace:
 [1] top-level scope
   @ REPL[13]:1
```


quinnj (commented):
Thanks!

PR #178, opened 2021-04-16
Implement C Data integration

quinnj:
This starts work towards supporting teh C data interface for the arrow
format, as documented
[here](https://arrow.apache.org/docs/format/CDataInterface.html#).

Currently in this PR, it includes struct definitions and basic
methods to allow getting a pointer to an `ArrowSchema`/`ArrowArray`
C-compatible struct that can then be populated by another
implementation. For example, with this PR, you can do:

```julia
using Arrow, PyCall
pd = pyimport("pandas")
pa = pyimport("pyarrow")
df = pd.DataFrame(py"""{'a': [1, 2, 3, 4, 5], 'b': ['a', 'b', 'c', 'd', 'e']}"""o)
rb = pa.record_batch(df)
sch = Arrow.CData.getschema() do ptr
    rb.schema._export_to_c(Int(ptr))
end
arr = Arrow.CData.getarray() do ptr
    rb._export_to_c(Int(ptr))
end
```

Currently, these `ArrowSchema`/`ArrowArray` structs are pretty bare
bones, but it at least lays some ground work for integration. Things we
still need/want to make all this nicer to use/work with:

  * Type format string parsing/converting: we need to parse the type
  format strings as outlined
  [here](https://arrow.apache.org/docs/format/CDataInterface.html#data-type-description-format-strings)
  to figure out what type of data we'll get in the arrays. It'd
  probably be best to add a `type` field to the ArrowSchema struct that
  we'd populate when converting from `CArrowSchema` -> `ArrowSchema`
  * Add a method like `Arrow.ArrowVector(::ArrowSchema, ::ArrowArray)`
  that produced a concrete `ArrowVector` subtype, like
  `Arrow.Primitive`, `Arrow.List`, etc. This will be a bit tricky,
  because have to follow all the same columnar layout trickery that we
  currently handle for IPC in the table.jl `build` methods. Perhaps we
  can refactor all that so we can re-use some code? Otherwise, we might
  just need to reimplement a bunch of that logic specific to converting
  `ArrrowArray`s.
  * That should give a robust consuming story; for producing, we
  probably need a definition like
  `Arrow.ArrowSchema(a::Arrow.ArrowVector)` that produced a valid
  `ArrowSchema`, and then overloads per `ArrowVector` subtype like
  `Arrow.ArrowArray(x::Arrow.Primitive)` that produced the right
  `ArrowArray` for a concrete arrow array
  * Then the last piece we need is just figuring out the right mechanics
  for providing a pointer to the `CArrowSchema`, `CArrowArray` structs
  once they're populated

If anyone would like to help out, I'm happy to provide as much guidance
as possible so others can get their feet wet in some arrow spec
nitty-gritty.

### DIFF
--- a/src/Arrow.jl
+++ b/src/Arrow.jl
@@ -92,6 +92,7 @@ include("arraytypes/arraytypes.jl")
 include("eltypes.jl")
 include("table.jl")
 include("write.jl")
+include("cinterface.jl")
 
 const LZ4_FRAME_COMPRESSOR = LZ4FrameCompressor[]
 const ZSTD_COMPRESSOR = ZstdCompressor[]
--- a/src/cinterface.jl
+++ b/src/cinterface.jl
@@ -1,0 +1,165 @@
+module CData
+
+export ArrowSchema, ArrowArray, getschema, getarray
+
+const ARROW_FLAG_DICTIONARY_ORDERED = 1
+const ARROW_FLAG_NULLABLE = 2
+const ARROW_FLAG_MAP_KEYS_SORTED = 4
+
+struct CArrowSchema
+    format::Ptr{UInt8}
+    name::Ptr{UInt8}
+    metadata::Ptr{UInt8}
+    flags::Int64
+    n_children::Int64
+    children::Ptr{Ptr{CArrowSchema}}
+    dictionary::Ptr{CArrowSchema}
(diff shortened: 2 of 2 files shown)


quinnj:
cc: @sa-, @Moelf

PR #179, opened 2021-04-17
WIP - C Data Interface

sa-:
Building off of this PR
https://github.com/JuliaData/Arrow.jl/pull/178

### DIFF
--- a/src/CDataInterface/format_string_to_types.jl
+++ b/src/CDataInterface/format_string_to_types.jl
@@ -1,0 +1,68 @@
+# https://arrow.apache.org/docs/format/CDataInterface.html#data-type-description-format-strings
+module FormatStrings
+
+function get_type_from_format_string(format_string ::String) ::DataType
+    if format_string == "n"
+        Nothing
+    elseif format_string == "b"
+        Bool
+    elseif format_string == "c"
+        Int8
+    elseif format_string == "C"
+        UInt8
+    elseif format_string == "s"
+        Int16
+    elseif format_string == "S"
+        UInt16
+    elseif format_string == "i"
+        Int32
+    elseif format_string == "I"
+        UInt32
+    elseif format_string == "l"
+        Int64
+    elseif format_string == "L"
+        UInt64
+    elseif format_string == "e"
+        Float16
+    elseif format_string == "f"
+        Float32
+    elseif format_string == "g"
+        Float64
+    elseif format_string == "z" || format_string == "Z"
+        Vector{UInt8}
+    elseif format_string == "u" || format_string == "U"
+        String
+    elseif format_string[1] == 'd'
+        const splits = split(format_string[3:end], ",")
+        precision = Int(splits[1])
+        scale = Int(splits[2])
+        if length(splits) == 3
+            bandwidth = splits[3]
+        end
+        #TODO return something here
+    elseif format_string[1] == 'w'
+        #TODO figure out fixed width binary
+    elseif format_string[1] == '+'
+        if format_string[2] == 'l' || format_string[2] == 'L'
+            Arrow.List
+        elseif format_string[2] == 'w'
+            size = Int(format_string[4:end]) #TODO use this somehow
+            Arrow.FixedSizeList
+        elseif format_string[2] == 's'
+            Arrow.Struct
+        elseif format_string[2] == 'm'
+            Arrow.Map
+        elseif format_string[2:3] == "ud"
+            type_strings = split(format_string[5:end], ",") # todo use this somehow
+            Arrow.DenseUnion
+        elseif format_string[2:3] == "us"
+            type_strings = split(format_string[5:end], ",") # todo use this somehow
+            Arrow.DenseUnion
+        end
+    elseif format_string[1] == 't'
+        if format_string[2:3]
+            Date
+    end
+end
+    
+end # module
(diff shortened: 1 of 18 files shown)


sa-:
```julia
using Arrow, PyCall
pd = pyimport("pandas")
pa = pyimport("pyarrow")
df = pd.DataFrame(py"""{'a': [1, 2, 3, 4, 5], 'b': ['a', 'b', 'c', 'd', 'e']}"""o)
rb = pa.record_batch(df)
sch = Arrow.CDataInterface.get_schema() do ptr
    rb.schema._export_to_c(Int(ptr))
end
arr = Arrow.CDataInterface.get_array() do ptr
    rb._export_to_c(Int(ptr))
end
```

src/CDataInterface/format_string_to_types.jl
```diff
@@ -0,0 +1,68 @@
+# https://arrow.apache.org/docs/format/CDataInterface.html#data-type-description-format-strings
+module FormatStrings
```
quinnj, line 2: This might be more of a personal style, but I wouldn't wrap this in a module here. In my mind, submodules are like mini packages that might be useful to split out on their own one day with stand-alone functionality. In this case, the format parsing stuff is pretty closely tied to the arrowschema, so I think it's better to just `include` the file.

src/CDataInterface/format_string_to_types.jl
```diff
@@ -0,0 +1,68 @@
...
+    elseif format_string == "z" || format_string == "Z"
+        Vector{UInt8}
+    elseif format_string == "u" || format_string == "U"
+        String
+    elseif format_string[1] == 'd'
+        const splits = split(format_string[3:end], ",")
+        precision = Int(splits[1])
+        scale = Int(splits[2])
+        if length(splits) == 3
+            bandwidth = splits[3]
```
quinnj, line 40: should be `bitwidth` instead of `bandwidth`

src/CDataInterface/format_string_to_types.jl
```diff
@@ -0,0 +1,68 @@
...
+        end
+        #TODO return something here
```
quinnj, line 42: In the `eltypes.jl` file, we define:
```julia
struct Decimal{P, S, T}
    value::T # only Int128 or Int256
end
```
which is what we should return here.

src/CDataInterface/format_string_to_types.jl
```diff
@@ -0,0 +1,68 @@
...
+    elseif format_string[1] == 'w'
+        #TODO figure out fixed width binary
```
quinnj, line 44: This will just be the same as `Arrow.FixedSizeList`, but with `UInt8` as the element type

quinnj: So a fixed width binary type won't have any children; it's like a hard-coded `UInt8` child type and so doesn't need to be parsed recursively.

src/CDataInterface/format_string_to_types.jl
```diff
@@ -0,0 +1,68 @@
...
+    elseif format_string[1] == '+'
+        if format_string[2] == 'l' || format_string[2] == 'L'
+            Arrow.List
```
quinnj, line 47: For these nested types, we'll need to parse the children recursively. So the overall method signature needs to take the full `ArrowSchema` type, and then access the format string at the top. Then when we get here, we'll call `get_type_from_format_string(sch.children[1])` and so on to get the List element type so we end up with a type like `Arrow.List{Vector{Int64}}` or whatever.

src/CDataInterface/format_string_to_types.jl
```diff
@@ -0,0 +1,68 @@
...
+        elseif format_string[2:3] == "ud"
+            type_strings = split(format_string[5:end], ",") # todo use this somehow
+            Arrow.DenseUnion
+        elseif format_string[2:3] == "us"
+            type_strings = split(format_string[5:end], ",") # todo use this somehow
+            Arrow.DenseUnion
+        end
+    elseif format_string[1] == 't'
+        if format_string[2:3]
+            Date
```
quinnj, line 64: The other thing we should probably support is a `convert::Bool=true` keyword arg to this function. It would allow the user to specify whether they'd like the native arrow type converted to a more natural Julia type or not. We support this in `Arrow.Table`. It's nice because I think there are some cases where the user just wants the raw arrow data type, but usually the user wants the nice Julia type. This is one of those cases where we have `Arrow.Date`, which is different from `Dates.Date`.

src/CDataInterface/format_string_to_types.jl
```diff
@@ -0,0 +1,99 @@
...
+                Arrow.Flatbuf.TimeUnitModule.SECOND
+            elseif format_string[3] == 'm'
+                Arrow.Flatbuf.TimeUnitModule.MILLISECOND
+            elseif format_string[3] == 'u'
+                Arrow.Flatbuf.TimeUnitModule.MICROSECOND
+            elseif format_string[3] == 'n'
+                Arrow.Flatbuf.TimeUnitModule.NANOSECOND
+            end
+
+            timezone = length(format_string) == 4 ? nothing : format_string[5:end]
```
sa-, line 90: Hey @quinnj , this line of code is incorrect. I couldn't quite figure out how to new up the timezone type. Is there any documentation around this?

sa-:
And in general, how does one new-up an ArrowVector? I was hoping to find a constructor that looks vaguely like this 
```
ArrowVector{T}(buffers ::Vector{ArrowBuffer{T}})
```

Issue #202, opened 2021-05-27
Error in "arrow json integration tests"


omus:
Noticed when running the CI for #201:
```
 arrow json integration tests: Error During Test at /home/runner/work/Arrow.jl/Arrow.jl/test/runtests.jl:59
  Got exception outside of a @test
  TaskFailedException
  Stacktrace:
    [1] wait
      @ ./task.jl:322 [inlined]
    [2] Arrow.Table(bytes::Vector{UInt8}, off::Int64, tlen::Nothing; convert::Bool)
      @ Arrow ~/work/Arrow.jl/Arrow.jl/src/table.jl:295
    [3] Arrow.Table(io::IOBuffer, pos::Int64, len::Nothing; kw::Base.Iterators.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:convert,), Tuple{Bool}}})
      @ Arrow ~/work/Arrow.jl/Arrow.jl/src/table.jl:208
    [4] macro expansion
      @ ~/work/Arrow.jl/Arrow.jl/test/runtests.jl:66 [inlined]
    [5] macro expansion
      @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
    [6] macro expansion
      @ ~/work/Arrow.jl/Arrow.jl/test/runtests.jl:61 [inlined]
    [7] macro expansion
      @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
    [8] top-level scope
      @ ~/work/Arrow.jl/Arrow.jl/test/runtests.jl:37
    [9] include(fname::String)
      @ Base.MainInclude ./client.jl:444
   [10] top-level scope
      @ none:6
   [11] eval
      @ ./boot.jl:360 [inlined]
   [12] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:261
   [13] _start()
      @ Base ./client.jl:485
  
      nested task error: BoundsError: attempt to access 0-element Vector{Int64} at index [0]
      Stacktrace:
       [1] getindex
         @ ./array.jl:801 [inlined]
       [2] append!(A::SentinelArrays.ChainedVector{Union{Missing, Bool}, Arrow.BoolVector{Union{Missing, Bool}}}, B::Arrow.BoolVector{Union{Missing, Bool}})
[...]

PR #204, opened 2021-05-27
Test expected log records

omus:
While investigating https://github.com/JuliaData/Arrow.jl/issues/202 I found the `@error` log to be distracting. We may as well test for expected log messages which also suppresses them.


omus:
Tests are currently expected to fail until #202 is fixed

omus:
Restarted CI now that #202 is fixed

quinnj (approved):
Thanks!

Issue #209, opened 2021-06-03
ArrowTypes as standalone package doesn't interoperate with Arrow.ArrowTypes


ericphanson:
I think Arrow should probably depend on ArrowTypes, not directly `include` it's code. The following shows that we have two separate modules with separate functions, e.g. `toarrow`:

```julia
julia> import Arrow

julia> import ArrowTypes

julia> Arrow.ArrowTypes
Arrow.ArrowTypes

julia> Arrow.ArrowTypes == ArrowTypes
false

julia> Arrow.ArrowTypes.toarrow == ArrowTypes.toarrow
false

julia> Arrow.ArrowTypes.toarrow() = 1

julia> methods(Arrow.ArrowTypes.toarrow)
# 12 methods for generic function "toarrow":
[1] toarrow() in Main at REPL[8]:1
[2] toarrow(::Nothing) in Arrow.ArrowTypes at /Users/eph/.julia/packages/Arrow/k23fl/src/ArrowTypes/src/ArrowTypes.jl:168
[3] toarrow(x::Char) in Arrow.ArrowTypes at /Users/eph/.julia/packages/Arrow/k23fl/src/ArrowTypes/src/ArrowTypes.jl:181
[4] toarrow(x::Symbol) in Arrow.ArrowTypes at /Users/eph/.julia/packages/Arrow/k23fl/src/ArrowTypes/src/ArrowTypes.jl:203
[5] toarrow(x::AbstractSet) in Arrow.ArrowTypes at /Users/eph/.julia/packages/Arrow/k23fl/src/ArrowTypes/src/ArrowTypes.jl:215
[6] toarrow(x::Base.UUID) in Arrow.ArrowTypes at /Users/eph/.julia/packages/Arrow/k23fl/src/ArrowTypes/src/ArrowTypes.jl:230
[7] toarrow(x::Dates.Date) in Arrow at /Users/eph/.julia/packages/Arrow/k23fl/src/eltypes.jl:208
[8] toarrow(x::Dates.Time) in Arrow at /Users/eph/.julia/packages/Arrow/k23fl/src/eltypes.jl:242
[9] toarrow(x::Dates.DateTime) in Arrow at /Users/eph/.julia/packages/Arrow/k23fl/src/eltypes.jl:291
[10] toarrow(x::TimeZones.ZonedDateTime) in Arrow at /Users/eph/.julia/packages/Arrow/k23fl/src/eltypes.jl:300
[...]

PR #212, opened 2021-06-14
Use standalone ArrowTypes package

omus:
Fixes: https://github.com/JuliaData/Arrow.jl/issues/209.

The one big hangup with this change is the inclusion of the `Manifest.toml`. The manifest allows use to dev `src/ArrowTypes` which means any changes to that code will immediately be reflected during local development. Additionally, since the manifest is committed the CI should also work the same way. Unfortunately, this can lead to changes to ArrowTypes.jl going unregistered (like what has currently happened).

Alternatively, we may want to not commit the Manifest.toml file and instead document how to setup a manifest file for local development and never commit it. This approach is more setup for developers but ensures that the CI is testing what users of the packages will ultimately end up seeing. I'm in favor of this approach


omus:
As 1.3 seems to be failing on the Manifest.toml I'll switch this over to using the other idea

quinnj:
Yes, I'm in favor of not committing the manifest file

omus:
Expecting CI to fail until #213 is merged and registered

.github/workflows/ci.yml
```diff
@@ -44,6 +44,44 @@ jobs:
...
+        version:
+          - '1.3'
+          - '1' # automatically expands to the latest stable 1.x release of Julia
+          - 'nightly'
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+    defaults:
+      working-directory: src/ArrowTypes
```
omus, line 62: Including the package name as part of the matrix would eliminate the YAML redundancy but this is the one part where doing that is hard. It may be possible to use symlinks to clean this up

omus: In fbbd33c19fcfa522de3cd6b58fbd33694935d85a I've switched over to using symlinks. I'm specifying `working-directory` on each step currently which has the best chance of working but I'll attempt to use `defaults` once I validate GHA works with the symlinks

Project.toml
```diff
@@ -17,6 +18,7 @@ TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
+ArrowTypes = "1.1"
```
omus, line 21: The Arrow.jl package should only requires ArrowTypes.jl 1.0 but the tests require 1.1.

omus:
@quinnj please approve the CI to run when you get the chance. There may be a little debugging to do on the CI side before this is ready

quinnj:
I don't have an option to approve CI; that would seem to indicate something in the ci.yml file that's causing CI not to run

ericphanson:
> something in the ci.yml file that's causing CI not to run 

If I edit it in the GitHub editor, it has syntax highlighting and shows errors:

Screenshot 2021-06-25 at 19 44 23

.github/workflows/ci.yml
omus: The `working-directory` solution [was a no-go](https://github.com/JuliaData/Arrow.jl/pull/212#issuecomment-868731166). Using `JULIA_PROJECT` now.

omus:
RTM. CI is now functional

codecov.yaml
```diff
@@ -0,0 +1,5 @@
+---
+codecov:
+  notify:
+    # Wait for all "test" matrix jobs
+    after_n_builds: 6
```
omus, line 5:
```suggestion
    after_n_builds: 6

```

README.md
```diff
@@ -27,6 +27,15 @@ or to use the official-apache code that follows the official apache release proc
 julia> using Pkg; Pkg.add(url="https://github.com/apache/arrow", subdir="julia/Arrow.jl")
 ```
 
+## Local Development
+
+When developing on Arrow.jl it is recommened that you run the following to ensure that any
```
ericphanson, line 32:
```suggestion
When developing on Arrow.jl it is recommended that you run the following to ensure that any
```

ericphanson (approved):

jrevels (approved):
LGTM. What are the follow ups for registration?

omus:
> LGTM. What are the follow ups for registration?

Should just be the registration. ArrowTypes version 1.1.0 has already [been registered](https://github.com/JuliaRegistries/General/pull/39003). Once this change is merged and registered any package which only depends on ArrowTypes.jl should be able to extend Arrow.jl without having to use Arrow.jl as a dependency.

omus:
I'm just validating that adding Arrow support through ArrowTypes.jl is fully working

omus:
> I'm just validating that adding Arrow support through ArrowTypes.jl is fully working

We're good here. My validation brought up some things I need to work through which are independent from these changes.

PR #223, opened 2021-07-08
Support Julia 1.0 for ArrowTypes package

omus:
Allows for packages to depend on ArrowTypes.jl and still include support for the Julia 1.0 LTS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
           - Arrow.jl
           - ArrowTypes.jl
         version:
+          - '1.0'
           - '1.3'
           - '1' # automatically expands to the latest stable 1.x release of Julia
           - 'nightly'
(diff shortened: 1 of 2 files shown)


omus:
The [Julia 1.0 failure](https://github.com/JuliaData/Arrow.jl/pull/223/checks?check_run_id=3015329136) is due to the build step using the Project.toml from the root of the directory while using Julia 1.0. That Project.toml wants to use ArrowTypes.jl 1.1 which requires Julia 1.3:
```
ERROR: Unsatisfiable requirements detected for package ArrowTypes [31f734f8]:
 ArrowTypes [31f734f8] log:
 ├─possible versions are: [1.0.0, 1.1.0] or uninstalled
 ├─restricted to versions 1.1.0-1 by an explicit requirement, leaving only versions 1.1.0
 └─restricted by julia compatibility requirements to versions: uninstalled — no versions left
```

jrevels (approved):

PR #224, opened 2021-07-08
Add `maxlog=1` to not spam logs

ericphanson:


quinnj (approved):
Thanks!

ericphanson:
I just realized that this might not log if there is more than 1 unsupported type:
```julia
julia> function f(x)
           @warn "Hello $x" x  maxlog=1
           return nothing
       end

julia> f(1)
┌ Warning: Hello 1
│   x = 1
└ @ Main REPL[10]:2

julia> f(20)

julia> f(1)

```
So actually maybe this was bad 😓

PR #225, opened 2021-07-09
use `_id` in warn logging

ericphanson:
Fixes the issue from https://github.com/JuliaData/Arrow.jl/pull/224#issuecomment-876772468:
```julia
julia> function f(x)
           @warn "Hello $x" x  maxlog=1 _id=hash((:f, x))
           return nothing
       end

julia> f(1)
┌ Warning: Hello 1
│   x = 1
└ @ Main REPL[24]:2

julia> f(1)

julia> f(1)

julia> f(2)
┌ Warning: Hello 2
│   x = 2
└ @ Main REPL[24]:2

julia> f(2)

julia>
```

It's really hard to add a test for this, because the Test logger doesn't respect `maxlog`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Arrow"
 uuid = "69666777-d1a9-59fb-9406-91d4454c9d45"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.6.0"
+version = "1.6.1"
 
 [deps]
 ArrowTypes = "31f734f8-188a-4ce0-8406-c8a06bd891cd"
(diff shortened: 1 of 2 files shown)


ericphanson:
I went ahead and bumped the patch so we can get this into a release 😄

ericphanson:
The `_id` argument is documented here, by the way: https://docs.julialang.org/en/v1/stdlib/Logging/#Logging.@logmsg

hannahilea:
lgtm! 👍

omus (approved):
LGTM. It would be great to have an issue filed to TestLogger not respecting `maxlog` if there isn't one already

ericphanson:
Filed! https://github.com/JuliaLang/julia/issues/41625

PR #229, opened 2021-07-23
custom struct getindex should deserialize based on the shape of the actual ArrowType, not the target JuliaType

jrevels:
This seems like it should be the case, and that `fromarrow` should be responsible for handling any fieldcount discrepancies, but I'm just messing about right now so could be missing something.

This arises in a case where the ArrowType has a different number of fields than the corresponding JuliaType

### DIFF
--- a/src/arraytypes/struct.jl
+++ b/src/arraytypes/struct.jl
@@ -33,20 +33,20 @@ isnamedtuple(T) = false
 istuple(::Type{<:Tuple}) = true
 istuple(T) = false
 
-@propagate_inbounds function Base.getindex(s::Struct{T}, i::Integer) where {T}
+@propagate_inbounds function Base.getindex(s::Struct{T,S}, i::Integer) where {T,S}
     @boundscheck checkbounds(s, i)
     NT = Base.nonmissingtype(T)
     if isnamedtuple(NT) || istuple(NT)
         if NT !== T
-            return s.validity[i] ? NT(ntuple(j->s.data[j][i], fieldcount(NT))) : missing
+            return s.validity[i] ? NT(ntuple(j->s.data[j][i], fieldcount(S))) : missing
         else
-            return NT(ntuple(j->s.data[j][i], fieldcount(NT)))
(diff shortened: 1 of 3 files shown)


src/arraytypes/struct.jl
```diff
@@ -33,20 +33,20 @@ isnamedtuple(T) = false
...
+@propagate_inbounds function Base.getindex(s::Struct{T,S}, i::Integer) where {T,S}
     @boundscheck checkbounds(s, i)
     NT = Base.nonmissingtype(T)
     if isnamedtuple(NT) || istuple(NT)
         if NT !== T
-            return s.validity[i] ? NT(ntuple(j->s.data[j][i], fieldcount(NT))) : missing
+            return s.validity[i] ? NT(ntuple(j->s.data[j][i], fieldcount(S))) : missing
         else
-            return NT(ntuple(j->s.data[j][i], fieldcount(NT)))
+            return NT(ntuple(j->s.data[j][i], fieldcount(S)))
```
omus, line 43: Not quite sure about this part. Giving it some thought

omus (commented):
Overall looks good. Solves the core issue in https://github.com/JuliaData/Arrow.jl/issues/134 (I'd like to fix the documentation before closing that though)

src/arraytypes/struct.jl
omus: If `S` contains more fields than `NT` we'll see an error here with the current code. The original code could also fail when the fieldcount of `NT` exceeded what was stored in `S`. As no one should be adding custom Arrow support for `Tuple` or `NamedTuple` subtypes I think this change is sensible to keep it consistent with the code below.

omus (approved):

src/arraytypes/struct.jl
omus: Aside: as far as I'm aware the `isnamedtuple(NT) || istuple(NT)` case here only exists as the constructors for `Tuple` and `NamedTuple` expect an iterator. We could probably handle this case using `fromarrow` with `@nospecialize` to simplify the logic here

PR #234, opened 2021-09-06
bugfix reading arrays

jonalm:
I have to admit that this is a bit above my head. But this fix prevents an alignment issue on reading array.

Compare to similar implementation at Flatbuffers.jl  https://github.com/JuliaData/FlatBuffers.jl/blob/e7d7ef0f842dab4d3793528061f3d5c35d05ba61/src/FlatBuffers.jl#L214


jrevels:
Hi @jonalm! Thanks for the contribution. Can you add a unit test?

quinnj:
Yes, sorry; @jonalm and I messaged on a slack a bit about this; we've been trying to come up with a test that would cover this case. The bug in question affects primitive arrays in flatbuffers, but, as far as I can tell, the only primitive array in the arrow spec is the unionIds array on the Union type object. And I'm discovering it's pretty hard to test that code. And recently saw a discussion in the broader arrow project about how that's not well-supported in other languages anyway.

jonalm:
I could generate a binary flatbuffer for which the parsing fails before the fix, generated by another language, and test whether Arrows.Flatbuffers parses that correctly. But in that case you kind of have to take my word for that the binary flatbuffer is correctly generated in the first place. Would that be acceptable?

jrevels:
Ah, thanks for the background.

> I could generate a binary flatbuffer for which the parsing fails before the fix, generated by another language, and test whether Arrows.Flatbuffers parses that correctly. But in that case you kind of have to take my word for that the binary flatbuffer is correctly generated in the first place. Would that be acceptable?

Yup, that seems like it might be the easiest path forward.

quinnj:
Yeah, that sounds good to me too.

jonalm:
Does that look okay @jrevels  @quinnj ?

jrevels:
Yup! confirmed test fails on `main`. Assuming CI goes green, seems good to merge to me 👍

jrevels (approved):

quinnj (approved):

Issue #211, opened 2021-06-13
use of `setmetadata!` on generic tables basically requires calling `Tables.columns` on the input


jrevels:
I thought there was an already issue for this, but couldn't find it:

```jl
julia> using Arrow, Tables

julia> t = Arrow.Table(Arrow.tobuffer((; x = [1,2,3], y = ["a", "b", "c"])))
Arrow.Table with 3 rows, 2 columns, and schema:
 :x  Int64
 :y  String

julia> Arrow.setmetadata!(t, Dict("key" => "value"))

julia> Arrow.getmetadata(t)
Dict{String,String} with 1 entry:
  "key" => "value"

# where did my metadata go?
julia> Arrow.getmetadata(Arrow.Table(Arrow.tobuffer(t))) isa Nothing
true

julia> t_cols = Tables.columns(t);

julia> Arrow.setmetadata!(t_cols, Dict("key" => "value"));

# but this works
julia> Arrow.getmetadata(Arrow.Table(Arrow.tobuffer(t_cols)))
Dict{String,String} with 1 entry:
  "key" => "value"
```

IIUC the reason for this is that `Arrow.write(_, t)` internally calls `Tables.columns(t)` and then tries to grab THAT object's metadata instead of the original `t`'s metadata (IIRC the exact place where this happens is [here](https://github.com/JuliaData/Arrow.jl/blob/b9cc6c1b4dc8b63c9ea340bbe63cb7bde6cb3242/src/write.jl#L138); I'd have to double check though, it's been awhile since I tracked it down).

haberdashPI:
If I understand this problem, what's ultimately needed is a change in Tables that provides support for metadata (e.g. by defining an interface for it that types can implement, or by transfering metadata to new objects upon a call to `Table.columns` etc... so that the new objects are also associated with the metadata in a global table).

ericphanson:
I think an alternate solution would be #90 -- just removing the metadata cache altogether, and only support it at read and write time. I think something like `table, metadata = Arrow.read("file.arrow")` and `Arrow.write("file.arrow", table, metadata)` would be simple and clear, and avoid these issues. Personally, I don't think Arrow should be in the business of trying to persist metadata through various Julia types and should just stick to (de)-serialization since there isn't a fully established metadata story in Julia's data ecosystem in general.

jrevels:
closed by #238

PR #238, opened 2021-09-12
remove global metadata cache, refactor custom_metadata API

### LINKED ISSUE #90: remove global metadata cache (OBJ_METADATA)
EDIT: I rescoped this issue, see https://github.com/JuliaData/Arrow.jl/issues/90#issuecomment-873094100

I noticed that metadata is stored in a global `IdDict` - would it make sense to provide an `unsetmetadata!(x)` (or use a sentinel e.g. `setmetadata!(x, nothing)`) that calls `delete!(OBJ_METADATA, x)`?

I could see a memory-leaky scenario where a e.g. a long running service writes a bunch of Arrow objects and attaches a small amount of metadata to each one and eventually OOMs or something.

jrevels:
essentially implements the approach described here https://github.com/JuliaData/Arrow.jl/issues/90#issuecomment-914870675

~I haven't run tests fully locally yet so may still have some lingering bugs.~ They pass 😎 

I also haven't been super performance-conscious here so we should be careful to make sure regressions aren't being introduced. 

Since we're doing a major bump for this anyway, I wonder whether it's a good time to fully delete some of the lingering deprecations here.

closes #90, #211

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -210,10 +210,4 @@ end
 
 toidict(x::Base.ImmutableDict) = x
 
-function toidict(pairs)
-    dict = Base.ImmutableDict(first(pairs))
-    for pair in Iterators.drop(pairs, 1)
-        dict = Base.ImmutableDict(dict, pair)
-    end
-    return dict
-end
+toiddict(pairs) = foldl(Base.ImmutableDict, pairs)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -210,4 +210,11 @@ end
 
 toidict(x::Base.ImmutableDict) = x
 
-toidict(pairs) = foldl(Base.ImmutableDict, pairs)
+# ref https://github.com/JuliaData/Arrow.jl/pull/238#issuecomment-919415809
+function toidict(pairs)
+    dict = Base.ImmutableDict(first(pairs))
+    for pair in Iterators.drop(pairs, 1)
+        dict = Base.ImmutableDict(dict, pair)
+    end
+    return dict
+end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -208,6 +208,8 @@ function tobuffer(data; kwargs...)
     return io
 end
 
+toidict(x::Base.ImmutableDict) = x
+
 function toidict(pairs)
     dict = Base.ImmutableDict(first(pairs))
     for pair in Iterators.drop(pairs, 1)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -207,3 +207,11 @@ function tobuffer(data; kwargs...)
     seekstart(io)
     return io
 end
+
+function toidict(pairs)
+    dict = Base.ImmutableDict(first(pairs))
+    for pair in Iterators.drop(pairs, 1)
+        dict = Base.ImmutableDict(dict, pair)
+    end
+    return dict
+end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -210,4 +210,4 @@ end
 
 toidict(x::Base.ImmutableDict) = x
 
-toiddict(pairs) = foldl(Base.ImmutableDict, pairs)
+toidict(pairs) = foldl(Base.ImmutableDict, pairs)
--- a/src/write.jl
+++ b/src/write.jl
@@ -37,7 +37,7 @@ By default, `Arrow.write` will use multiple threads to write multiple
 record batches simultaneously (e.g. if julia is started with `julia -t 8` or the `JULIA_NUM_THREADS` environment variable is set).
 
 Supported keyword arguments to `Arrow.write` include:
-  * `colmetadata=nothing`: TODO
+  * `colmetadata=nothing`: the metadata that should be written as the table's columns' `custom_metadata` fields; must either be `nothing` or an `AbstractDict` of `column_name::Symbol => column_metadata` where `column_metadata` is an iterable of `<:AbstractString` pairs.
   * `compress`: possible values include `:lz4`, `:zstd`, or your own initialized `LZ4FrameCompressor` or `ZstdCompressor` objects; will cause all buffers in each record batch to use the respective compression encoding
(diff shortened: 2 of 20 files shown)


jrevels:
Flipping out of draft as tests pass and docs are updated (unless I missed anything).

Before merging, should we decide on:

> Since we're doing a major bump for this anyway, I wonder whether it's a good time to fully delete some of the lingering deprecations here.

We also could do this in a separate PR before tagging

src/write.jl
```diff
@@ -229,26 +198,28 @@ end
...
-        newcol = toarrowvector(col, i, dictencodings, dictencodingdeltas; compression=compress, largelists=largelists, denseunions=denseunions, dictencode=dictencode, dictencodenested=dictencodenested, maxdepth=maxdepth)
+        oldcolmeta = getmetadata(col)
+        newcolmeta = isnothing(colmeta) ? oldcolmeta : get(colmeta, nm, oldcolmeta)
+        newcol = toarrowvector(col, i, dictencodings, dictencodingdeltas, newcolmeta; compression=compress, largelists=largelists, denseunions=denseunions, dictencode=dictencode, dictencodenested=dictencodenested, maxdepth=maxdepth)
         newtypes[i] = eltype(newcol)
         newcols[i] = newcol
     end
     minlen, maxlen = isempty(newcols) ? (0, 0) : extrema(length, newcols)
     minlen == maxlen || throw(ArgumentError("columns with unequal lengths detected: $minlen < $maxlen"))
+    meta = isnothing(meta) ? nothing : toidict(String(k) => String(v) for (k, v) in meta)
```
ericphanson, line 222:
```suggestion
    meta = _normalizemeta(meta)
```

ericphanson (commented):
exciting! Didn't know about `Base.ImmutableDict`, cool.

src/utils.jl
```diff
@@ -207,3 +207,13 @@ function tobuffer(data; kwargs...)
...
+
+toidict(x::Base.ImmutableDict) = x
+
+function toidict(pairs)
+    dict = Base.ImmutableDict(first(pairs))
+    for pair in Iterators.drop(pairs, 1)
+        dict = Base.ImmutableDict(dict, pair)
+    end
+    return dict
+end
```
quinnj, line 219:
```suggestion
toiddict(pairs) = foldl(Base.ImmutableDict, pairs)
```

quinnj (approved):
LGTM; thanks for taking this on! I left one suggestion to simplify `toiddict`.

src/utils.jl
jrevels: actually had something like this originally but seemed like `foldl` was a bit slower than the more manual loop (didn't investigate why):

```jl
julia> pairs = [randstring(30) => randstring(30) for _ in 1:15];

julia> f(x) = foldl(Base.ImmutableDict, x)

julia> @benchmark f($pairs)
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.442 μs … 179.375 μs  ┊ GC (min … max): 0.00% … 98.91%
 Time  (median):     1.479 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.541 μs ±   3.071 μs  ┊ GC (mean ± σ):  3.44% ±  1.71%

       ▃ ▇ █ ▅ ▃ ▅ ▅ ▃ ▄ ▄ ▁
  ▂▃▃▇▁█▁█▁█▁█▁█▁█▁█▁█▁█▁█▁█▁█▁▇▁▇▁▆▁▅▁▅▁▄▁▄▁▄▁▃▁▃▁▂▁▂▁▂▁▂▁▂▂ ▄
  1.44 μs         Histogram: frequency by time        1.57 μs <

 Memory estimate: 1.66 KiB, allocs estimate: 60.

julia> @benchmark toidict($pairs)
BenchmarkTools.Trial: 10000 samples with 964 evaluations.
 Range (min … max):   90.076 ns …   1.652 μs  ┊ GC (min … max):  0.00% … 93.47%
 Time  (median):     102.524 ns               ┊ GC (median):     0.00%
 Time  (mean ± σ):   117.682 ns ± 152.728 ns  ┊ GC (mean ± σ):  13.49% ±  9.63%

  █▂                                                            ▁
  ██▄▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅ █
  90.1 ns       Histogram: log(frequency) by time       1.57 μs <
```

Maybe the magnitude here doesn't matter that much in reality though...you'd have to have small arrow messages and lots of metadata, which doesn't seem like a likely/good case to optimize for...

so let's go with the simpler `foldl` as you suggest and can optimize later if it ever matters

src/utils.jl
```diff
@@ -207,3 +207,7 @@ function tobuffer(data; kwargs...)
     seekstart(io)
     return io
 end
+
+toidict(x::Base.ImmutableDict) = x
+
+toiddict(pairs) = foldl(Base.ImmutableDict, pairs)
```
jrevels, line 213:
```suggestion
toidict(pairs) = foldl(Base.ImmutableDict, pairs)
```

fix typo

jrevels:
From Slack convo: 

I'll merge this PR and can open a follow-up later to delete the deprecated regions before we actually tag the v2.0 release.

jrevels:
Hmm. Looks like Julia 1.3 tests now fail because the `foldl` implementation of `toidict` relies on `Base.ImmutableDict(::Pair, ::Pair)` working, which I guess isn't defined on Julia 1.3

and the Julia 1.6 tests fail because `foldl` doesn't apply the given `op` on unary inputs (which is expected):  

```jl
julia> foldl(Base.ImmutableDict, ("a" => "b",))
"a" => "b"
```

I think I'll just revert back to the original more verbose implementation

PR #241, opened 2021-09-20
delete deprecations in preparation for v2.0 release

jrevels:
Follow-up to #238. 

Unsure whether I've hit everything we'd want to cover here, but at least hits the items that I remembered/stood out in the code to me.


quinnj (approved):
LGTM

PR #246, opened 2021-09-28
Add ability to pass directory of inputs to Arrow.Table/Arrow.Stream

quinnj:
Implements #235. The proposed implementation here isn't too complicated;
it introduces a new `ArrowBlob` type that we'll convert any
`IO`/`String`/`Vector{UInt8}` input to, and the main
`Arrow.Table`/`Arrow.Stream` methods take a `Vector{ArrowBlob}` to
operate on. The schema of each input must match in order to be treated
as a single "table". "Must match" in this PR means `==(sch1, sch2)`,
which does include `custom_metadata` and dictionaries, which might not
be desireable, but I haven't quite had the chance to think through all
the way yet. Otherwise, I think the rest here is pretty straightfoward
and non-breaking.

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -14,6 +14,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+struct ArrowBlob
+    bytes::Vector{UInt8}
+    pos::Int
+    len::Int
+end
+
+ArrowBlob(bytes::Vector{UInt8}, pos::Int, len::Nothing) = ArrowBlob(bytes, pos, length(bytes))
+
+tobytes(bytes::Vector{UInt8}) = bytes
+tobytes(io::IO) = Base.read(io)
+function tobytes(str)
+    f = string(str)
+    isfile(f) || throw(ArgumentError("$f is not a file"))
+    return Mmap.mmap(f)
+end
+
 struct BatchIterator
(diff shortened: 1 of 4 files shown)


src/table.jl
```diff
@@ -54,21 +72,22 @@ end
...
-    if len > 24 &&
-        _startswith(bytes, off, FILE_FORMAT_MAGIC_BYTES) &&
-        _endswith(bytes, off + len - 1, FILE_FORMAT_MAGIC_BYTES)
-        off += 8 # skip past magic bytes + padding
-    end
+function Stream(inputs::Vector{ArrowBlob}; convert::Bool=true)
     dictencodings = Dict{Int64, DictEncoding}() # dictionary id => DictEncoding
     dictencoded = Dict{Int64, Meta.Field}() # dictionary id => field
-    batchiterator = BatchIterator(bytes, off)
+    blob = inputs[1]
```
layne-sadler, line 85: why pass a vector if it's just going to be accessed by index?

quinnj: I mean, there isnt' really an `Indexable` interface to "loosen" the type restriction here, so a `Vector` is a natural candidate. We could probably loosen to `AbstractArray`, but that's an easy change if someone needs to pass something other than an array.

Issue #266, opened 2021-12-16
Performance regression caused by Arrow type piracy


NHDaly:
Hi! :) 👋 

We observed a mysterious performance regression in some benchmarks when we started using `Arrow` in our codebase, and we have narrowed it down to type piracy on `getproperty` of Types.

Here's an MRE:
```julia
julia> using BenchmarkTools

julia> @btime sort!([rand([Int,Float64,Int8,Int32]) for _ in 1:100], by=fieldcount);
  49.939 μs (103 allocations: 12.33 KiB)

julia> using Arrow

julia> @btime sort!([rand([Int,Float64,Int8,Int32]) for _ in 1:100], by=fieldcount);
  687.034 μs (103 allocations: 12.33 KiB)
```

The regression seems to be that calling `getproperty` on types is now a dynamic dispatch, whereas it was not previously. I believe this is caused by the following method, which prevents julia from devirtualizing the getproperty calls elsewhere:
https://github.com/JuliaData/Arrow.jl/blob/8d486c05cf441e78b4764f8a67b9ca903e88dafb/src/FlatBuffers/FlatBuffers.jl#L138

You can see that effect here:
```julia
julia> methods(getproperty, (Type, Symbol))
# 1 method for generic function "getproperty":
[1] getproperty(x::Type, f::Symbol) in Base at Base.jl:28

julia> using Arrow

julia> methods(getproperty, (Type, Symbol))
# 11 methods for generic function "getproperty":
[1] getproperty(::Type{Arrow.Flatbuf.DateUnitModule.DateUnit}, sym::Symbol) in Arrow.Flatbuf.DateUnitModule at /Users/nathandaly/.julia/dev/Arrow/src/FlatBuffers/FlatBuffers.jl
:138
[2] getproperty(::Type{Arrow.Flatbuf.BodyCompressionMethodModule.BodyCompressionMethod}, sym::Symbol) in Arrow.Flatbuf.BodyCompressionMethodModule at /Users/nathandaly/.julia/dev/Arrow/src/FlatBuffers/FlatBuffers.jl:138
[3] getproperty(::Type{Arrow.Flatbuf.CompressionTypeModule.CompressionType}, sym::Symbol) in Arrow.Flatbuf.CompressionTypeModule at /Users/nathandaly/.julia/dev/Arrow/src/FlatBuffers/FlatBuffers.jl:138
[...]

Issue #265, opened 2021-12-07
Transfer repository to apache organization


quinnj:
As has been discussed in a few places (slack, arrow-dev mailing list, etc.), we'd like to transfer this repo to the apache organization in our efforts to officially participate in the broader arrow community and project. For the mechanics of the transfer, @kou has very helpfully laid out the plan and TODO items [here](https://lists.apache.org/thread/cr28lsyq1rsxsb810olsj74h7z08zvbg) and we can track the current state of the transfer [here](https://incubator.apache.org/ip-clearance/arrow-julia-library2.html).

There is a request to *NOT* merge any more commits until the transfer is complete; cc active committers: @omus, @jrevels.

To complete the transfer, we just need signed CLAs for the following individuals:

- [x] @jrevels 
- [x] @omus 
- [x] @ericphanson 
- [x] @dmbates 

This list is as computed by @kou for more-than-1-commit since the original donation of the repo to the arrow project. The required CLA can be found [here](https://www.apache.org/licenses/contributor-agreements.html). @jrevels , @omus , @ericphanson, as also noted, if your collective contributions should be viewed under a "corporate" contribution, there's a separate corporate CLA. Happy to chat more if that doesn't make sense or whatever.

Let's do this!

jrevels:
> The required CLA can be found here. @jrevels , @omus , @ericphanson, as also noted, if your collective contributions should be viewed under a "corporate" contribution, there's a separate corporate CLA. 

I don't believe we were thoroughly tracking this, but I think the safest assumption would be to assume that out of our contributions, at least some were made on behalf of Beacon/in the course of Beacon work.

In that case, does it make the most sense to basically just do both? i.e. @jrevels / @omus / @ericphanson all provide individual CLAs, and I also execute a corporate CLA for Beacon?

dmbates:
Do we simply use "arrow" in the notify project box?  The agreement states that we email the signed PDF only to secretary@apache.org Is there any further action that is needed for coordination of the submission?

quinnj:
I think it's fine to use `arrow` or `arrow-julia` as the project box, preferring `arrow`. Yes, you only need to email and then if you'd check your name off hte list above that'd be great.

@jrevels, it's probably safest to do both then.

ericphanson:
I’ve submitted my ICLA, can someone check me off the list? (I don’t have permission)

quinnj:
@jrevels, @omus, @KristofferC; any update?

omus:
[...]

PR #267, opened 2021-12-17
Proposal: change `@scopedenum` to make modules to avoid type piracy

NHDaly:
This changes the result of `@scopedenum` to produce a module instead of trying to stick the enums into the enum type, as a proposed fix for https://github.com/JuliaData/Arrow.jl/issues/266.

It's admittedly less pretty, but I think it's still pretty nice! It seems like a hopefully somewhat understandable pattern to follow, which is that it pluralizes your enum name to make a module. We communicate that to the users by returning it at the end of the expression:
```julia
julia> Arrow.FlatBuffers.@scopedenum MyEnum X=1 Y=2
Main.MyEnums

julia> MyEnums.X
X::MyEnum = 1
```

I can see why this is less appealing than what was there before, but what is there before suffers from the performance problems identified in https://github.com/JuliaData/Arrow.jl/issues/266.

Thanks for your consideration! ❤️


NHDaly:
Verifying this fixes https://github.com/JuliaData/Arrow.jl/pull/267:
```julia
julia> using BenchmarkTools

julia> @btime sort!([rand([Int,Float64,Int8,Int32]) for _ in 1:100], by=fieldcount);
  59.247 μs (103 allocations: 12.33 KiB)

julia> using Arrow
[ Info: Precompiling Arrow [69666777-d1a9-59fb-9406-91d4454c9d45]

julia> @btime sort!([rand([Int,Float64,Int8,Int32]) for _ in 1:100], by=fieldcount);
  53.731 μs (103 allocations: 12.33 KiB)
```

ericphanson:
btw @NHDaly this looks like a great change to me (although I'm not very familiar with the code in question), but FYI commits are frozen for now while they transfer the repo to apache's org (ref https://github.com/JuliaData/Arrow.jl/issues/265)

NHDaly:
:o oh shoot, thanks @ericphanson! Very interesting. Makes sense.

Thanks. So, should we wait until all that is done? Do you know how long that will take? In the meantime, I think we'll do our prototyping against my fork from which i made this PR, which should keep us unblocked for now i guess...?

ericphanson:
Maybe @quinnj knows about the timeline? I don't know myself, sorry.

Developing against the fork makes sense to me. As an alternative, if it will be awhile, I have an idea that I have never tried but I think might work. If you have a private registry you could register your fork there (with the same name + uuid). The way the package manager works is it will take versions from any install registry. So you could do something like register patches like `v"2.2.1"` , `v"2.2.2"`, etc from your fork into your private registry, and then ask for the next release of this Arrow.jl to be `v"2.3.0"`. So when that comes out, the version would be later than your fork and would be used instead. I.e. basically inserting some extra versions into the list. I am not a maintainer/committer here so I can't promise that the next version would be `v"2.3.0"` but maybe the maintainers here could accomodate that if you want to go this route.

NHDaly:
Hah, yeah cool. For better or worse we just have one big monorepo with only one manifest, so it's not a problem for us to build against the fork, but thanks, that's a super interesting idea if we ever changed to that approach!! ♥️ thanks for all the info Eric, this is helpful.

Hope you've been doing well!! ☺️☺️

quinnj:
It shouldn't be too long; I think we're just waiting on @jrevels to sign the CLA and then we can make the transfer.

NHDaly:
Cool, thanks Jacob! :) Lemme know if you want to chat about this at all when you're ready. 👍

jrevels:
xref https://github.com/JuliaData/Arrow.jl/issues/265#issuecomment-998993831, hopefully we're one step closer to unblocking this :) happy to see that Relational is also making use of Arrow.jl!

quinnj (dismissed):
Hmmm......yeah, I guess this is ok. I don't love the magical "text transformation" to make it plural, but I'm not sure there are many other better ideas. I had wanted to eventually split this out into a separate package or something, but perhaps we should just make a push to put in Base and be able to resolve the performance problems there perhaps.

quinnj:
@kou, can you approve the tests running here and if passing, merge please?

kou:
@quinnj Sure! But #273 blocks running CI. Please wait for a response from INFRA.

quinnj:
Ah yes, I forgot those aren't enabled yet. Ok, we'll wait.

NHDaly:
> I don't love the magical "text transformation" to make it plural, but I'm not sure there are many other better ideas.

One better text transformation which would be more reliable than the pluralization could be to rename `@enum Foo` to `FooModule` or `FooEnums` if you think you might prefer it? I agree this is a bit stinky.

But thanks for the approval, yeah, i didn't think of any better alternatives either. 👍

kou (approved):

kou:
CI is runnable again.

kou (approved):
+1

CI is green.

NHDaly:
Hooray. Thanks!

How will the release process work under the new setup? Should I have bumped a version number in my PR? Or will you do that separately afterwards?

Issue #295, opened 2022-03-04
Order of record batches from "arrow file" format files (i.e. `Arrow.Table`) not preserved


complyue:
https://github.com/apache/arrow-julia/blob/614fce0a5d7db8fee078be32690c5220848538e2/src/table.jl#L276-L293

I see from above that record batches will be parsed (esp. decompression could be rather intensive computation workload) in parallel if the Julia runtime has multithread enabled, which is great. 

But according to the implementation, the original order of batches as they had been written will not be guaranteed as preserved, which I think is not ideal. I'm not sure how Arrow spec should say about this aspect, but I'm dealing with time series data recorded batch-by-batch where the order signifies a lot.

I'd like to draft a PR to preserve batch order with regard to this concern, and as I start tinkering with the codebase, I file this issue to ask your opinions about it.

(Btw, I'm also tinkering about a PR for #293, which is orthogonal wrt functionality, but seems closely related wrt implementation details. I'd think 2 separate PRs would make better clarity for review and release purpose, but if you can accept a single PR addressing the 2 things together, it could be a lot easier for me, given I'm not fluent in git rebasing and related skills.)

PR #296, opened 2022-03-04
Solving #295

complyue:
Attempt to solve #295, please review this PR.


quinnj:
@complyue, thanks for the contribution! I spent some time reviewing your PR and something seemed very familiar about the problem. And that's when I remembered that we already solved this on the `Arrow.write` side, so I think it makes sense to use the same utility type (`OrderedChannel`) on the reading side. See my PR here: https://github.com/apache/arrow-julia/pull/298

complyue:
@quinnj, thanks for providing your solution, I prefer your PR instead of this one, so closing.

I've also removed changes in this PR from PR #297, so #297 becomes a clean PR on its own, please review that sooner, thanks!

Issue #293, opened 2022-03-02
Support Tables.partitions when reading "arrow file" format in addition to "stream" format


complyue:
I see #23 closed but only *stream* format got supported. 

How about files in "arrow file" format, each with multiple record batches inside?

With released version I get a single `Table` as the result of `Tables.partitions(tbl)` is a `::Tuple{Arrow.Table}`, however the columns there are chunked.

Can we get partition tables with contiguous column arrays via `Tables.partitions`?

---

As for workarounds before that's supported, any public API I can use to obtain contiguous column arrays per each record batches?

quinnj:
PR up: https://github.com/apache/arrow-julia/pull/443

PR #297, opened 2022-03-04
Solving #293

complyue:
Attempt to solve #293, implementing `Tables.partitions()` for `Arrow.Table` turns out trivial enough, though I'm not clear what to do about companion test cases, documentation etc.

I'm lazy to have not excluded #296 from this PR, I'm open to draft another PR separated if you'd like, but the change set seems small, I hope you can have a quick look and decide what to do next.

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -286,7 +286,7 @@ function Tables.partitions(t::Table)
     end
     return (t,)
   end
-  
+
 # high-level user API functions
 Table(input, pos::Integer=1, len=nothing; kw...) = Table([ArrowBlob(tobytes(input), pos, len)]; kw...)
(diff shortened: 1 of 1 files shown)


complyue:
With an alternative solution #298, changes from #296 are removed from this PR, now it's only about `Arrow.Table`'s `Tables.partitions()` implementation.

I suggest the commit history should be squashed, but don't have experience how to do that from my side, only seeing articles suggesting "Squash and merge" option of Github, but that seems only doable when merging a PR.

quinnj:
Sorry for never getting to review/support this. I agree the functionality is good and this was a good start. Instead of returning a tuple of tables, I opted for a more lazy approach of just defining a custom iterator + tests: https://github.com/apache/arrow-julia/pull/443

PR #298, opened 2022-03-06
Alternative fix for #295.

quinnj:
We already have a utility defined (`OrderedChannel`) that we use when
writing record batches to ensure batches get _written_ in the same order
they are provided; it makes sense to use the same utility when reading
to ensure incoming record batches are _read_ in the appropriate order.


baumgold (approved):

quinnj:
Closing in favor of https://github.com/apache/arrow-julia/pull/339

Issue #232, opened 2021-08-12
Serializing `Dict{String,Real}` result in garbage values 


omus:
Serializing a `Dict` which contains `Bool` and `Float64` values results in a Arrow generating garbage values:
```julia
julia> d = Dict("is_valid" => true,"probability" => 0.53495216)
Dict{String, Real} with 2 entries:
  "is_valid"    => true
  "probability" => 0.534952

julia> t = Arrow.Table(Arrow.tobuffer((; value=[d])))
Arrow.Table with 1 rows, 1 columns, and schema:
 :value  Dict{String, Float64}

julia> t.value
1-element Arrow.Map{Dict{String, Float64}, Int32, Arrow.Struct{NamedTuple{(:key, :value), Tuple{String, Float64}}, Tuple{Arrow.List{String, Int32, Vector{UInt8}}, Arrow.Primitive{Float64, Vector{Float64}}}}}:
 Dict("is_valid" => -6.6622794774424345e159, "probability" => 3.1e-322)
```

Note that pre-converting the values to `Float64` doesn't result in this behaviour:
```julia
julia> d = Dict{String,Float64}("is_valid" => true,"probability" => 0.53495216)
Dict{String, Float64} with 2 entries:
  "is_valid"    => 1.0
  "probability" => 0.534952

julia> t = Arrow.Table(Arrow.tobuffer((; value=[d])))
Arrow.Table with 1 rows, 1 columns, and schema:
 :value  Dict{String, Float64}

julia> t.value
1-element Arrow.Map{Dict{String, Float64}, Int32, Arrow.Struct{NamedTuple{(:key, :value), Tuple{String, Float64}}, Tuple{Arrow.List{String, Int32, Vector{UInt8}}, Arrow.Primitive{Float64, Vector{Float64}}}}}:
 Dict("is_valid" => 1.0, "probability" => 0.53495216)
```

quinnj:
Yeah, similar to what we do w/ arrays, we should probably try to enforce the Dict valtype with the `concrete_or_concreteunion` machinery in ArrowTypes.jl. Or we at least need a check in map.jl that it's a concrete type/union when serializing.

FuZhiyu:
I noticed this has been fixed by the PR above. Will it be included in the release any time soon?

ericphanson:
The fix from #305 was included in Arrow.jl v2.3+. I'll close this issue then, unless folks think there is more that should be done in this case.

PR #305, opened 2022-03-15
disallow non-concrete map-like types to prevent incorrect serialization

jrevels:
Potential fix for #232, implements the simple check suggested by @quinnj in that thread.

This is technically breaking - the following works on `main` just fine, but will error after this PR due to the check:

```jl
julia> Arrow.Table(Arrow.tobuffer((; x = [Dict{Real,Real}(1 => 2)])))
Arrow.Table with 1 rows, 1 columns, and schema:
 :x  Dict{Int64, Int64}
```

IMO dropping support for nonconcrete keytypes/valtypes is preferable to the silent failure, but it would be good if we can have our cake and eat it too (which I think @quinnj 's other suggestion was referring to). 

@quinnj , when you mentioned:

> similar to what we do w/ arrays, we should probably try to enforce the Dict valtype with the concrete_or_concreteunion machinery in ArrowTypes.jl.

did you mean basically implementing a `Dict`-like version of https://github.com/apache/arrow-julia/blob/f88a62e0b6458ed6ea0439fc371f2e4ee0e039a7/src/ArrowTypes/src/ArrowTypes.jl#L336-L357, and wrapping incoming map-like values in that?

I can take a stab at that too, if so, but might not get time to do so this week; in that case, it might be worth folding this in earlier (since the silent failure mode here is pretty bad, IMO) and then recovering the functionality in a subsequent release. Up to you, though.

### DIFF
--- a/src/arraytypes/map.jl
+++ b/src/arraytypes/map.jl
@@ -51,7 +51,10 @@ function arrowvector(::MapKind, x, i, nl, fi, de, ded, meta; largelists::Bool=fa
     validity = ValidityBitmap(x)
     ET = eltype(x)
     DT = Base.nonmissingtype(ET)
-    KT = KeyValue{keytype(DT), valtype(DT)}
+    KDT, VDT = keytype(DT), valtype(DT)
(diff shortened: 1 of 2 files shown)


quinnj (approved):
Yeah, that was my idea: make a Dict version of what we do for arrays. I think we should do this for now though.

PR #309, opened 2022-03-17
Change scoped enums behavior

jonalm:
Implement suggestions for https://github.com/apache/arrow-julia/issues/308


quinnj (commented):
Yeah, I like this approach; the failing tests are probably because we need to update the usages in the Arrow package itself

jonalm:
I will to find time for that within a couple of days. But before I do so, should we reconsider the name of the primitive type, i.e. something other than with leading underscore? I just saw the announcement of enumx.jl https://discourse.julialang.org/t/ann-enumx-jl-improved-enums-for-julia/78096 , which has implemented a similar idea, but using `T` as the name of the type inside the module. E.g. for `@scopedenum MyEnum::Base.UInt16  A=0 B=1` the values would be accessed as `MyEnum.A` and the primitive type could be accessed as `MyEnum.T`.

jonalm:
On second thought that `T` idea brakes down (without careful handleing) for enums containing `T` fields, which I guess are way more likely than `_<enumname>`.

quinnj (approved):
Yes, I think this is a good correction; thanks @jonalm. The `T.__TYPE__` isn't ideal, but I think it's a decent solution.

quinnj:
I went ahead and rebased

jonalm:
@quinnj  Is there something blocking getting this into main?

ararslan:
Would it be possible to replace the custom implementation of scoped enums here with a dependency on [EnumX](https://github.com/fredrikekre/EnumX.jl)? That has no dependencies and is very small.

quinnj:
Sorry all that this languished; I've since become more familiar w/ the EnumX.jl package and agree it's the right way to go. I'll try to do a PR soon switching over to use it. Sorry @jonalm for all the headache here, but thanks for all the help along the way!

PR #321, opened 2022-04-29
fix version mismatch by changing footer to V5

### LINKED ISSUE #320: Versions in footer and message do not agree, this causes issues reading Arrow files with other libraries (such as `arrow-rs`).
here version is set as V5
https://github.com/apache/arrow-julia/blob/a3f6da7c1d59f8321315f4955a3b45c48d38aab4/src/write.jl#L374

and here as V4
https://github.com/apache/arrow-julia/blob/a3f6da7c1d59f8321315f4955a3b45c48d38aab4/src/write.jl#L374

This causes issues reading files written with Arrow.jl from other languages' arrow implementations.

pcjentsch:
fixes #320

hopefully V5 is the correct version for both of these

With apache/arrow-rs#1631, files written by Arrow.jl are readable by `arrow-rs`.

### DIFF
--- a/src/write.jl
+++ b/src/write.jl
@@ -125,8 +125,8 @@ mutable struct Writer{T<:IO}
     msgs::OrderedChannel{Message}
     schema::Ref{Tables.Schema}
     firstcols::Ref{Any}
-    dictencodings::Dict{Int64, Any}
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Thanks @pcjentsch!

PR #322, opened 2022-05-07
Fix repo name in the Readme

CarloLucibello:


kou:
Good catch!
But we can remove the "Difference between this code and the apache/arrow/julia/Arrow repository" section entirely because we've moved to here from apache/arrow/julia/Arrow. Could you remove the section entirely?

CarloLucibello:
Done! Side note: I find a little confusing that this repo is not named Arrow.jl contrary to usual Julia naming convention and since it provides the package Arrow.jl, but no big deal

kou (approved):
Thanks!

> Side note: I find a little confusing that this repo is not named Arrow.jl contrary to usual Julia naming convention and since it provides the package Arrow.jl, but no big deal

`apache/*` uses `apache/#{project}-#{subproject}` naming convention. For this repository, `project` is `arrow` and `subproject` is `julia`.

PR #325, opened 2022-06-10
allow ntasks to be 0 and determine whetheher to use threads based on nthreads rather than ntasks

baumgold:
This allows configuring an unbuffered channel while still using multithreading

### DIFF
--- a/src/write.jl
+++ b/src/write.jl
@@ -46,7 +46,7 @@ Supported keyword arguments to `Arrow.write` include:
   * `largelists::Bool=false`: causes list column types to be written with Int64 offset arrays; mainly for testing purposes; by default, Int64 offsets will be used only if needed
   * `maxdepth::Int=$DEFAULT_MAX_DEPTH`: deepest allowed nested serialization level; this is provided by default to prevent accidental infinite recursion with mutually recursive data structures
   * `metadata=Arrow.getmetadata(tbl)`: the metadata that should be written as the table's schema's `custom_metadata` field; must either be `nothing` or an iterable of `<:AbstractString` pairs.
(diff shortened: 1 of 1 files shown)


quinnj (approved):

src/write.jl
```diff
@@ -135,16 +135,13 @@ mutable struct Writer{T<:IO}
...
-        throw(ArgumentError("ntasks keyword argument must be > 0; pass `ntasks=1` to disable multithreaded writing"))
-    end
     msgs = OrderedChannel{Message}(ntasks)
     schema = Ref{Tables.Schema}()
     firstcols = Ref{Any}()
     dictencodings = Dict{Int64,Any}() # Lockable{DictEncoding}
     blocks = (Block[], Block[])
     # start message writing from channel
-    threaded = ntasks > 1
+    threaded = Threads.nthreads() > 1
```
ericphanson, line 144: isn't this breaking? since someone could've been relying on "to disable multithreaded writing, pass `ntasks=1`" if they are writing to some `io` handle that doesn't use locking internally. (I wasn't so maybe it's theoretical, just looks like a strange change to me)

baumgold: Thanks for your comment.  I don't think this change is breaking.  In this context, `threaded` just determines whether to use threads or coroutines.  In either case the processing of each partition is done in parallel and then sequenced through an `OrderedChannel` to be serialized to the IO.  The IO serialization is sequential (not parallel) since order matters so no locking is required.  We should use threads when they are available, otherwise we should use coroutines.

The `ntasks` parameter is orthogonal to the choice of threads or coroutines.  `ntasks` defines how large the buffer can be within the `OrderedChannel`.  A larger buffer means the parallel processing of partitions can outpace  the sequential serialization of the results to the IO since the processed results can be cached in the buffer.

ericphanson: ah ok, thank you for clarifying, that seems reasonable & non-breaking to me too.

PR #338, opened 2022-10-05
support reading empty files that are in the process of being written

baumgold:
Currently trying to read an empty file that is in the process of being written throws an exception.  A cleaner way to handle this is to return an empty table/stream.

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -58,15 +58,26 @@ lazily converted to more friendly Julia defaults; by default, `convert=true`.
 mutable struct Stream
     inputs::Vector{ArrowBlob}
     inputindex::Int
-    batchiterator::BatchIterator
-    pos::Int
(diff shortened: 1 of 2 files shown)


baumgold:
@quinnj - would you mind reviewing/approving? Thanks.

quinnj (approved):
LGTM; thanks.

baumgold:
@quinnj - can you trigger a release including this fix? Thanks.

quinnj:
@baumgold, will you make a PR incrementing the Project.toml version and we'll rope @kou in to help with that?

kou:
@baumgold FYI:
* https://github.com/apache/arrow-julia/blob/main/dev/release/README.md#overview : Our release process
* https://github.com/apache/arrow-julia/pull/312 : The PR for the previous release.

PR #339, opened 2022-10-06
Use OrderedSynchronizer instead of OrderedChannel

quinnj:


kou (approved):
+1

How about referring #295 when we merge this?

quinnj:
Thanks @kou

PR #343, opened 2022-10-10
Remove requirement to have 1 approval on PRs

quinnj:
See discussion [here](https://github.com/apache/arrow-julia/issues/342#issuecomment-1272411069).


kou (approved):
+1

PR #349, opened 2022-10-28
fix Julia 1.6

SimonDanisch:
When I look at the CI jobs on the last commit on `main` it looks like Arrow tests aren't run against `1.6`, but the CI.yml looks like it does consider 1.6 - but it's also a more complex setup, I guess something may go wrong somewhere?
Anyways, the latest tagged version of Arrow.jl fails to compile on Julia 1.6...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,14 +79,6 @@ jobs:
           - ubuntu-latest
         arch:
           - x64
-        exclude:
-          # Test Arrow.jl/ArrowTypes.jl on their oldest supported Julia versions
-          - pkg:
-              name: Arrow.jl
-            version: '1.6'
-          - pkg:
-              name: ArrowTypes.jl
-            version: '1.4'
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,7 @@
 name = "Arrow"
 uuid = "69666777-d1a9-59fb-9406-91d4454c9d45"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "2.4.0"
+version = "2.4.1"
 
 [deps]
 ArrowTypes = "31f734f8-188a-4ce0-8406-c8a06bd891cd"
--- a/src/arraytypes/dictencoding.jl
+++ b/src/arraytypes/dictencoding.jl
@@ -128,7 +128,7 @@ function arrowvector(::DictEncodedKind, x::DictEncoded, i, nl, fi, de, ded, meta
         de[id] = Lockable(x.encoding)
     else
         encodinglockable = de[id]
-        @lock encodinglockable begin
+        Base.@lock encodinglockable begin
             encoding = encodinglockable.value
             # in this case, we just need to check if any values in our local pool need to be delta dicationary serialized
             deltas = setdiff(x.encoding, encoding)
@@ -195,7 +195,7 @@ function arrowvector(::DictEncodedKind, x, i, nl, fi, de, ded, meta; dictencode:
           # if value doesn't exist in encoding, push! it
           # also add to deltas updates
         encodinglockable = de[id]
-        @lock encodinglockable begin
+        Base.@lock encodinglockable begin
             encoding = encodinglockable.value
             len = length(x)
             ET = indextype(encoding)


bkamins:
@quinnj - is this line
https://github.com/apache/arrow-julia/blob/b49255ad486fce9d8fe6f30370583de5bbffd895/.github/workflows/ci.yml#L86
disabling Arrow.jl tests on Julia 1.6?

bkamins:
The reason for error is that `@lock` is not exported in Julia 1.6. Its export was added in https://github.com/JuliaLang/julia/pull/39588

ericphanson:
Looking at the git history, I think those exclusions were originally intended so that ArrowTypes and Arrow could have different minimum supported Julia versions (xref https://github.com/apache/arrow-julia/pull/223), but were incorrectly updated over time, such that now they incorrectly exclude things.

ericphanson:
Can you remove those exclusions from the workflow and bump the patch version @SimonDanisch? Then we can see CI passing on 1.6 here and merge

quinnj:
Oof, yeah, sorry about the mixup here folks.

ericphanson:
cc @kou apparently Arrow.jl v2.4 is totally broken on Julia 1.6 and won't even load, causing CI issues in a lot of places (particularly private repos at Beacon where we use a lot of Arrow.jl). Can we either:
* expedite the patch release faster than 24 hr, OR
* register this in the Julia package registry before the official release vote goes through?

ericphanson (approved):

ericphanson:
> cc @kou apparently Arrow.jl v2.4 is totally broken on Julia 1.6 and won't even load, causing CI issues in a lot of places (particularly private repos at Beacon where we use a lot of Arrow.jl). Can we either:
> 
> * expedite the patch release faster than 24 hr, OR
> * register this in the Julia package registry before the official release vote goes through?

Actually, we can just yank the 2.4.0 release, meaning it won't be installed by default anymore. This will solve the CI issues. Then we can land 2.4.1, waiting the 24 hours.

ericphanson:
Ok, 2.4.0 is yanked on the General registry, so at least we don't have the broken release causing CI problems. Now we can re-land the release as 2.4.1.

bkamins (approved):

bkamins:
We might want to wait with the release for https://github.com/apache/arrow-julia/pull/350 to be merged (that PR is small, but requires a review from someone who has good understanding of different requirements of host platforms)

bkamins:
(but if that PR is problematic I would not wait with the release)

ericphanson:
I tried making a 2.4.1-rc1 but I don't have java installed, nor a PGP key, and I am out of motivation for today

baumgold:
It seems we're still not able to make a 2.4 release?  I lost track of what the holdup is at this point...

ericphanson:
Someone needs to do the steps from https://github.com/apache/arrow-julia/tree/main/dev/release#release

kou:
> 1. Test the revision to be released

Could you confirm that the current main is releasable?

> 2. Increment version number in `Project.toml`

Could someone open a pull request for it?

Then, I can do the following tasks:

> 3. Prepare RC and vote (detailed later)
> 4. Publish (detailed later)

If there is a person who wants to try these tasks, I can support the person.

ericphanson:
This PR incremented the version number, in anticipation of a release. Is it better next time if that is a separate PR?

kou:
Oh, I didn't notice it.
I think that a separate PR is better because we need to choose which version (major/minor/patch) to be incremented when we put multiple changes to one release.
But we can choose this mixed approach by discussing and reaching a consensus. It's for "Open Communications" in [the Apache way](anticipation). If it's needed, we should open a new issue for it.
 
Could you confirm that the current main doesn't include a backward compatibility change nor a new feature? (I think that this project follows semantic versioning, right?)

ericphanson:
> this project follows semantic versioning 
Yep!

> confirm that the current main doesn't include a backward compatibility change nor a new feature?  

Since v2.4.0 we have
* this PR: bugfix
* https://github.com/apache/arrow-julia/pull/356: bugfix
* https://github.com/apache/arrow-julia/pull/355: internal tooling change / not really sure if it's a bugfix or feature
* https://github.com/apache/arrow-julia/pull/357: bugfix
* https://github.com/apache/arrow-julia/pull/358: not sure, bugfix or feature

Maybe someone else can weigh in about #355 and #358

quinnj:
All bugfixes; a patch release should be appropriate

kou:
Vote thread: https://lists.apache.org/thread/mp8mg70c01kvs4648ttq1kq4vq9nr7cv

kou:
Passed: https://lists.apache.org/thread/q9bjmktlowkj5xm93nsw5cpjh2kgmjtf

Could someone update the Julia General registry?

ericphanson:
Ok, I've made https://github.com/JuliaRegistries/General/pull/72443 to register 2.4.1. When that PR is merged, it will be registered in General

PR #351, opened 2022-10-28
Add `svn` as a release dependency

ericphanson:
I get:
```
dev/release/release_rc.sh: line 80: svn: command not found
```


ericphanson:
Ah sorry, this isn't right, the release docs already mention SVN, and this is for verify

PR #355, opened 2022-10-31
replace custom debug handling with LoggingExtras

baumgold:
[LoggingExtras.jl](https://github.com/JuliaLogging/LoggingExtras.jl) already handles this functionality.


nickrobinson251:
the standard Julia logging doesn't quite support the same thing as this, as here we have `@debug 1 msg` and `@debug 2 message` to set different levels within the "debug" level. But @quinnj did move similar functionality into LoggingExtras.jl (https://github.com/JuliaLogging/LoggingExtras.jl/pull/73), so we could remove this code here and use `@debugv` from LoggingExtras.jl to achieve the same thing

quinnj:
Yes, that was my plan eventually (switch to the debugv macro in LoggingExtras).

baumgold:
Hmm I guess I mistook Julia's "Programmatically defined levels" feature:

https://github.com/JuliaLang/julia/blob/v1.8.2/base/logging.jl#L183
https://github.com/JuliaLang/julia/blob/v1.8.2/test/corelogging.jl#L36-L41

It seems like this ability to log as non-standard levels and filter loggers for non-standard levels should be supported in Julia's Base.  I'll read @quinnj's LoggingExtras.  Thanks.

quinnj (approved):
This LGTM for now. Thanks for the cleanup @baumgold!

The other step I'd like to do here is that at the top level of `Arrow.Stream`/`Arrow.Table`/`Arrow.write`/`Arrow.append` (i.e. main user entrypoints), to add a keyword argument `debug::Int=0` where the user could simply pass `debug=2` and then we'd wrap the body of the methods in:

```julia
LoggingExtras.withlevel(Debug; verbosity=debug) do
    # method body
end
```
I feel like that's a super convenient way to quickly turn on/off the debug logs without having to import LoggingExtras yourself or get the withlevel syntax right.

We don't need to do that here if you'd rather not; I can add it later.

baumgold:
I'm not a big fan of integrating log-level en/disabling logic within the Arrow APIs as it should really be an orthogonal concept.  I understand that your proposal make enabling debug logging convenient, but it also seems to add some extra complexity to an already complex codebase.  Maybe your proposal is actually very clean and elegant but I don't see how yet.  I'll keep thinking about it but for now I think we should add that change in a separate branch/PR.

PR #356, opened 2022-11-01
fix BatchIterator iterate method to handle partial messages

baumgold:
The current implementation blindly assumes the message is complete and can cause undefined behavior if it is not.  This fixes the iterate method to ignore partial messages.

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -418,10 +418,18 @@ function Base.iterate(x::BatchIterator, (pos, id)=(x.startpos, 0))
         return nothing
     end
     pos += 4
+    if pos + msglen - 1 > length(x.bytes)
+        @debug 1 "not enough bytes left to read Meta.Message"
+        return nothing
+    end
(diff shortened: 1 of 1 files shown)


quinnj (approved):
This seems good to me. It's a little more flexible/loose than throwing an error, but I think it should be fine.

baumgold:
@quinnj - thanks for reviewing/approving.  I still don't have write access (working on getting permissions from Apache).  In the mean time would you mind merging for me, please?  I'd like to propose a new release (again).

kou:
@baumgold Could you see my e-mail reply for your Apache account? You need to configure https://gitbox.apache.org/ .

baumgold:
> @baumgold Could you see my e-mail reply for your Apache account? You need to configure https://gitbox.apache.org/ .

Got it.  Configured gitbox.  All is well with the world.  Many thanks.

Issue #61, opened 2020-11-05
Releasing the lock on arrow file


bkamins:
When you read the file using `Arrow.Table` then a lock on the file is made. I could not find a function to release the lock within a session on Windows.
The only way that works is doing:
1. removing all references to the table.
2. doing `GC.gc()`
3. then you can remove the file

This is valid and makes sense, but the question is if there is some other way to do it (maybe not).

If not then I understand that the recommended pattern for such cases is to open file, pass IO to `ArrowTable`, and close file. Is this correct and recommended pattern if you want to avoid memory mapping and locking the file?

This is needed in the scenarios when you want to use Arrow as a temporary storage on disk but want to clean up things after the session.

(I am writing a blog post on Arrow.jl now)

quinnj:
Yes, that is currently the correct way to release the file lock for windows (due to mmaping). We could add a finalizer on `Arrow.Table` that also finalized the underlying buffer, which would mean you could just do `finalize(table)` and the lock would be released (though the `table` would then be invalid). I can look into doing that.

bkamins:
Thank you. I think there is no rush with the finalizer. I just wanted to make sure that I correctly understand the current design.

PR #361, opened 2022-11-15
don't run tests that fail on Windows

visr:
This is a bit crude, but there are currently many IOErrors on Windows, coming from calling `Arrow.Table(path::String)` where path is in a temp dir that gets cleaned up at the end of `mktemp`/`mktempdir`.

```julia
mktempdir() do dir
    path = joinpath(dir, "test.arrow")
    Arrow.write(path, t)
    t2 = Arrow.Table(path)
end
```

One can avoid this by using

```julia
open(path) do io
    Arrow.Table(io)
end
```

But then we're not testing the Mmap and String methods, so I'm simple disabling these tests on Windows. Related to #61, which perhaps should be reopened, to see if we can find a resolution.

With this all tests pass on Windows except for the `resize!` issue mentioned in https://github.com/apache/arrow-julia/pull/357#issuecomment-1301943392.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -45,17 +45,18 @@ end
 end # @testset "table roundtrips"
 
 @testset "table append" begin
+    # skip windows since file locking prevents mktemp cleanup
+    if !Sys.iswindows()
+        for case in testtables
+            testappend(case...)
+        end
 
-    for case in testtables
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Yeah, this sounds simple enough; thanks.

PR #365, opened 2022-11-28
Ensure elements are converted when indexed from ArrowTypes.ToArrow

### LINKED ISSUE #364: PooledArray are incorrectly saved
If a float-type element in ```PooledVector{Real, UInt32, Vector{UInt32}}``` is replaced by an integer, Arrow will incorrectly save this value.

```
using Arrow, DataFrames, PooledArrays

df = DataFrame(x = PooledArray(Vector{Real}([1.0])))
df[1, 1] = 2
Arrow.write("test.arrow", df)
Arrow.Table("test.arrow") |> DataFrame
```

The above code incorrectly saves 2 as 1.0e-323.

```
1×1 DataFrame
 Row │ x        
     │ Float64
─────┼──────────
   1 │ 1.0e-323
```

quinnj:
Fixes #364. This is an unfortunate bug that was sneaking in under the guise of "supposedly passing" tests. i.e. we had a test like:

```julia
@test ArrowTypes.ToArrow(Any[1, 3.14]) == [1.0, 3.14]
```
_BUT_, this was misleading, because, in reality, we had:

```julia
x = ArrowTypes.ToArrow(Any[1, 3.14])
@test x === 1
```

i.e. when you indexed an element of `ToArrow`, no actual conversion was happening. So even though `ToArrow` was doing the work of figuring out the right concrete promoted type or union type, when you indexed later, it was just giving you the original element. Which actually works in a number of cases, for example, if the promoted type was a `Union` and all elements were one of the Union types, but in this case, where the promoted case was `Float64` and one element was an integer, we just got the plain integer out, when in reality, we want that integer converted to a float.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -37,7 +37,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 WorkerUtilities = "76eceee3-57b5-4d4a-8e66-0e911cebbf60"
 
 [compat]
-ArrowTypes = "1.1"
+ArrowTypes = "1.1,2"
 BitIntegers = "0.2"
 CodecLz4 = "0.4"
 CodecZstd = "0.7"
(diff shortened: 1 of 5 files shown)


jrevels (approved):
❤️

quinnj:
So one thing I noticed w/ this PR is that our Arrow.jl compat actually didnt' allow ArrowTypes 2.0, so I bumped that and now we're getting some test failures. I can't quite remember what broke in ArrowTypes 2.0, but I'll look into it a bit later tonight.

PR #369, opened 2022-12-26
Create CompatHelper.yml

CarloLucibello:


.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,45 @@
+name: CompatHelper
```
quinnj, line 1: One of the CI checks is failing because we have to include the apache license text at the top of every single file (like [here](https://github.com/apache/arrow-julia/blob/main/.github/workflows/TagBot.yml)).

quinnj:
@CarloLucibello, can you remind me what teh CompatHelper does for the repo? I can't quite remember what it's used for.

CarloLucibello:
It automatically files a PR updating the compact bounds (only when needed) whenever a new release of one of the dependencies is out

quinnj (approved):
Thanks!

Issue #367, opened 2022-12-23
inappropriately applied `Arrow.NullVector` optimization to `Union{ZonedDateTime,Missing}` column


jrevels:
It seems like `Union{ZonedDateTime,Missing}` (de)serialization can bork Arrow.jl in some circumstances, presumably to a misapplied `Arrow.NullVector` optimization (or code in that optimization assuming a constructor exists):

```jl
julia> using Arrow, TimeZones

julia> t = (; x=Union{ZonedDateTime,Missing}[missing])
(x = Union{Missing, ZonedDateTime}[missing],)

julia> a = Arrow.Table(Arrow.tobuffer(t))
Arrow.Table with 1 rows, 1 columns, and schema:
 :x  ZonedDateTime

julia> a.x[1]
ERROR: MethodError: no method matching ZonedDateTime(::Missing)
Closest candidates are:
  ZonedDateTime(::Union{Dates.Period, TimeZone}...) at ~/.julia/packages/TimeZones/2oXvO/src/types/zoneddatetime.jl:136
  ZonedDateTime(::AbstractString) at ~/.julia/packages/TimeZones/2oXvO/src/parse.jl:86
  ZonedDateTime(::AbstractString, ::AbstractString; locale) at ~/.julia/packages/TimeZones/2oXvO/src/parse.jl:100
  ...
Stacktrace:
 [1] fromarrow(#unused#::Type{ZonedDateTime}, x::Missing)
   @ ArrowTypes ~/.julia/packages/ArrowTypes/dkiHE/src/ArrowTypes.jl:157
 [2] getindex(v::Arrow.NullVector{ZonedDateTime}, i::Int64)
   @ Arrow ~/.julia/packages/Arrow/QsQ3U/src/arraytypes/arraytypes.jl:99
 [3] top-level scope
   @ REPL[13]:1
```

Note that if we insert an actual `ZonedDateTime` and avoid the `NullVector` optimization:

```jl
julia> t = (; x=Union{ZonedDateTime,Missing}[missing, ZonedDateTime("0001-01-01T00:00:00.000Z")])
(x = Union{Missing, ZonedDateTime}[missing, ZonedDateTime(1, 1, 1, tz"Z")],)

julia> a = Arrow.Table(Arrow.tobuffer(t))
Arrow.Table with 2 rows, 1 columns, and schema:
 :x  Union{Missing, ZonedDateTime}

julia> a.x[1]
missing

julia> a.x[2]
[...]

PR #371, opened 2023-01-04
Fix issue with missing and non-concrete Arrow types

omus:
Fixes https://github.com/apache/arrow-julia/issues/367

### DIFF
--- a/src/ArrowTypes/Project.toml
+++ b/src/ArrowTypes/Project.toml
@@ -18,7 +18,7 @@
 name = "ArrowTypes"
 uuid = "31f734f8-188a-4ce0-8406-c8a06bd891cd"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "2.0.0"
+version = "2.1.0"
 
 
 [deps]
--- a/src/ArrowTypes/src/ArrowTypes.jl
+++ b/src/ArrowTypes/src/ArrowTypes.jl
@@ -344,6 +344,9 @@ function ToArrow(x::A) where {A}
         for i = 2:length(x)
             @inbounds T = promoteunion(T, typeof(toarrow(x[i])))
         end
+        if T === Missing
+            T = promoteunion(T, typeof(toarrow(default(S))))
+        end
     end
     return ToArrow{T, A}(x)
 end
--- a/src/ArrowTypes/test/tests.jl
+++ b/src/ArrowTypes/test/tests.jl
@@ -146,4 +146,23 @@ x = ArrowTypes.ToArrow(Any[1, 3.14])
 @test x[2] === 3.14
 @test ArrowTypes.ToArrow(Any[1, 3.14, "hey"]) == [1.0, 3.14, "hey"]
 
+@testset "respect non-missing type" begin
+    struct DateTimeTZ
+        instant::Int64
+        tz::String
+    end
+
+    struct Timestamp{TZ}
+        x::Int64
+    end
+
+    ArrowTypes.ArrowType(::Type{DateTimeTZ}) = Timestamp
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -67,6 +67,9 @@ jobs:
       fail-fast: false
       matrix:
         pkg:
+          - name: Arrow.jl (dev'd ArrowTypes.jl)
+            dir: '.'
+            dev: './src/ArrowTypes'
           - name: Arrow.jl
             dir: '.'
```
omus, line 74: For this PR the tests will only pass when using the local ArrowTypes.jl. However, we probably don't want to have this always be the way we run tests as otherwise we're not testing that the ArrowTypes compat entry is correct. This is a compromise which requires the PR reviewer to be aware of the scenarios in which it's okay for some of these checks to fail.

I happy to remove these changes from these PR or make them into a separate PR. I included them here to demonstrate that the Arrow.jl tests are functional with the latest ArrowTypes.jl

omus:
@quinnj would love to get some feedback on this PR

src/ArrowTypes/src/ArrowTypes.jl
```diff
@@ -344,6 +344,9 @@ function ToArrow(x::A) where {A}
         for i = 2:length(x)
             @inbounds T = promoteunion(T, typeof(toarrow(x[i])))
         end
+        if T === Missing
```
quinnj, line 347: So there's an interesting case to consider here: `Any[missing, missing]`. The array eltype is `Any`, so we hit this codepath, and `T` is `Missing`, so we'll hit this new branch, so uh-oh, that might do something bad, right? Well, `ArrowTypes.default(Any)` returns `nothing`, and `toarrow(nothing)` returns.......`missing`! So we end up with the right `T` in the end, albeit somewhat unexpectedly? I would only worry a bit if we end up changing `toarrow(::Nothing) = missing`.

For now, can we just add a test for this case and then deal w/ it later if necessary?

quinnj (approved):
Thanks @omus!

quinnj:
It looks like the "verify release" check is failing, @omus is that what you were referring to above where the arrow tests will fail because they're not using the arrowtypes local/latest?

omus:
> It looks like the "verify release" check is failing, @omus is that what you were referring to above where the arrow tests will fail because they're not using the arrowtypes local/latest?

Yup, this is the issue to which I was referring. Seems like the best path forward is to just include the ArrowTypes.jl changes in this PR and keep the actual tests for the change in a follow up PR.

src/ArrowTypes/src/ArrowTypes.jl
omus: Sounds good. I'll try to take care of this and the other outstanding things this week.

.github/workflows/ci.yml
omus: Removed the CI changes

src/ArrowTypes/Project.toml
```diff
@@ -18,8 +18,7 @@
 name = "ArrowTypes"
 uuid = "31f734f8-188a-4ce0-8406-c8a06bd891cd"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "2.0.1"
-
+version = "2.0.2"
```
omus, line 21: I considered making this a minor release due to `ArrowTypes.default` not being fully documented

omus:
Good to proceed from my end. Just hoping for a final review

omus:
I just thought of a good way to add the tests for #367 into this PR. Will make one more change here

quinnj (approved):

src/ArrowTypes/Project.toml
ericphanson: We haven’t registered 2.0.1 yet btw, so automerge will complain, but that’s ok and we can manually merge

PR #382, opened 2023-01-20
Get dictionaries of children only when field.children not nothing 

### LINKED ISSUE #375: Make recursive iteration to get dictionaries more defensive for interop
The problem I have encountered is due to this for loop:

https://github.com/apache/arrow-julia/blob/528c3ba9eaa38b7e9c1ff3d4c998f5b304ec64cf/src/table.jl#L387

It assumes that `field.children` is an iterable and, to my understanding, this assumption is usually warranted as for the last elements, when the leaves of the tree are reached, `field.children` is `Arrow.Flatbuf.Field[]`. However, we have observed that when Arrow files are written in C# `field.children === nothing` and trying to iterate throws an error. I can share an example file.

I could fix this in my `dev`ed, local Arrow.jl by simply conditioning the for loop:

```julia
if field.children !== nothing
    for child in field.children
        getdictionaries!(dictencoded, child)
    end
end
```

I am not sure if this is the right fix but I am opening this issue here because I am more familiar with the Arrow implementation in Julia and I think it is reasonable to be more defensive. But, as reading the same data works once the file is imported-exported with Python/pyarrow, my guess is that it is the C# implementation of Arrow that is doing something fishy here. Maybe this can also be fixed at the site where Julia produces the `nothing` when translating the Arrow types?

okartal:
Closes #375 

This PR fixes a problem in reading an Arrow IPC file generated by C# Apache.Arrow 10.0.1 where the last recursive iteration to get the dictionaries returns `nothing` instead of an empty iterable.

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -384,8 +384,10 @@ function getdictionaries!(dictencoded, field)
     if d !== nothing
         dictencoded[d.id] = field
     end
-    for child in field.children
-        getdictionaries!(dictencoded, child)
+    if field.children !== nothing
+        for child in field.children
+            getdictionaries!(dictencoded, child)
+        end
     end
     return
 end


okartal:
@quinnj I want to attach a test file that produces the following error
image
but Github says it is not allowed

okartal:
Is there another way I can share the arrow test file?

quinnj:
@okartal, so sorry for the very slow response here. The best thing would be to include the example file in your commits in this PR. Then we can just add it as a test that this is working. If you just want to commit the file, I can help actually writing a test that reads the file.

quinnj (approved):
This LGTM, though it makes me slightly worried if there are other cases where we may want to do extra validation of not `nothing` from flatbuffer fields.

PR #390, opened 2023-02-09
add IPv4/IPv6 support to ArrowTypes

baumgold:


quinnj (approved):
Sorry for the slow review here @baumgold! LGTM

PR #400, opened 2023-03-12
add kwarg chunksize for default data partitioning for write

svilupp:
This PR proposes to introduce automated partitioning of the provided tables when writing.

Nowadays, most machines are multithreaded and `Arrow.write()` provides multithreaded writing for partitioned data. However, a user must explicitly partition their data. 
Unfortunately, most users do not realize that both their write and subsequent read operations will not be multithreaded without such partitioning (there is an issue to improve the docs).

This PR defaults to partitioning data if it's larger than 64K rows (should be beneficial on most systems) to enable better Arrow.jl performance on both read and write.

Implementation:
- the new kwarg is called `chunksize` (maps to PyArrow and should be broadly understood)
- uses default `chunksize` of 64000 rows, as per [`PyArrow.write_feather`](https://arrow.apache.org/docs/python/generated/pyarrow.feather.write_feather.html)
- allows users to opt-out by providing `chunksize=nothing`
- partitioning is done via `Iterators.partition(Tables.rows(tbl),chunksize)` for all Tables.jl-compatible sources (checks `Tables.istable`)

Some resources:
- [PyArrow write_feather docs](https://arrow.apache.org/docs/python/generated/pyarrow.feather.write_feather.html)
- Dataframes.jl introduces overload for `Iterators.partition` in [1.5 Release](https://github.com/JuliaData/DataFrames.jl/pull/3212)
- [Arrow.jl author's blog post on partioning](https://quinnj.hashnode.dev/partition-all-the-datas) (sidenote: I really enjoy your posts - please write more :-) )

### DIFF
--- a/src/write.jl
+++ b/src/write.jl
@@ -48,14 +48,21 @@ Supported keyword arguments to `Arrow.write` include:
   * `metadata=Arrow.getmetadata(tbl)`: the metadata that should be written as the table's schema's `custom_metadata` field; must either be `nothing` or an iterable of `<:AbstractString` pairs.
   * `ntasks::Int`: number of buffered threaded tasks to allow while writing input partitions out as arrow record batches; default is no limit; for unbuffered writing, pass `ntasks=0`
   * `file::Bool=false`: if a an `io` argument is being written to, passing `file=true` will cause the arrow file format to be written instead of just IPC streaming
+  * `chunksize::Union{Nothing,Integer}=64000`: if a table is being written, this will cause the table to be partitioned into chunks of the given size (`chunksize` rows); if `nothing`, no partitioning will occur
 """
 function write end
 
 write(io_or_file; kw...) = x -> write(io_or_file, x; kw...)
 
-function write(file_path, tbl; kwargs...)
+function write(file_path, tbl; chunksize::Union{Nothing,Integer}=64000, kwargs...)
+    if !isnothing(chunksize) && Tables.istable(tbl)
+        @assert chunksize >= 0 "chunksize must be >= 0"
+        tbl_source = Iterators.partition(Tables.rows(tbl),chunksize)
+    else
+        tbl_source = tbl 
+    end
     open(Writer, file_path; file=true, kwargs...) do writer
-        write(writer, tbl)
+        write(writer, tbl_source)
     end
     file_path
 end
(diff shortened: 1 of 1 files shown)


svilupp:
I've changed the condition for automatic partitioning to be `Tables.rowaccess()=true` as well, to prevent accepting some columntables without row iterators.

In addition, I've changed `Iterators.partition(Tables.rows(tbl),chunksize)` to `Iterators.partition(tbl,chunksize)` to avoid missingness type getting lost (eg, for DataFrames)

svilupp:
Added compat for DataFrames via Extras

src/write.jl
```diff
@@ -48,14 +48,29 @@ Supported keyword arguments to `Arrow.write` include:
...
   * `ntasks::Int`: number of buffered threaded tasks to allow while writing input partitions out as arrow record batches; default is no limit; for unbuffered writing, pass `ntasks=0`
   * `file::Bool=false`: if a an `io` argument is being written to, passing `file=true` will cause the arrow file format to be written instead of just IPC streaming
+  * `chunksize::Union{Nothing,Integer}=64000`: if a table is being written, this will cause the table to be partitioned into chunks of the given size (`chunksize` rows); if `nothing`, no partitioning will occur
 """
 function write end
 
 write(io_or_file; kw...) = x -> write(io_or_file, x; kw...)
 
-function write(file_path, tbl; kwargs...)
+function write(file_path, tbl; chunksize::Union{Nothing,Integer}=64000, kwargs...)
```
baumgold, line 57: I think `chunksize` should move to be a new field  in `Writer` with default kwarg value set in the `Base.open` constructor on L170.  This would eliminate the code duplication.

src/write.jl
```diff
@@ -278,9 +293,23 @@ function Base.close(writer::Writer)
     nothing
 end
 
-function write(io::IO, tbl; kwargs...)
+function write(io::IO, tbl; chunksize::Union{Nothing,Integer}=64000, kwargs...)
+    # rowaccces is a necessary pre-requisite for row-iteration (not sufficient though)
+    if !isnothing(chunksize) && Tables.istable(tbl) && Tables.rowaccess(tbl)
+        @assert chunksize >= 0 "chunksize must be >= 0"
+        if hasmethod(Iterators.partition,(typeof(tbl),))
+            tbl_source = Iterators.partition(tbl, chunksize)
```
baumgold, line 301: Can we use `Iterators.partition` from Base rather than `DataFrames` to prevent adding one more dependency?

https://docs.julialang.org/en/v1/base/iterators/#Base.Iterators.partition

baumgold:
> By far the simplest option would be to pass the schema down - because we have access to it before `Tables.columns` is called in the arrow construction (that's how we lose the schema, because we "materialize" the chunk as is)

We could allow users to optionally provide the Schema in the Base.open constructor of the Writer object.  If a user makes use of this then we should validate the the actual schema of each chunk matches that of the expected schema.

PR #404, opened 2023-03-14
fix Base.eltype methods and functions that take Type parameters

baumgold:


src/ArrowTypes/src/ArrowTypes.jl
```diff
@@ -205,7 +205,7 @@ struct ListKind{stringtype} <: ArrowKind end
 
 ListKind() = ListKind{false}()
 isstringtype(::ListKind{stringtype}) where {stringtype} = stringtype
-isstringtype(::Type{ListKind{stringtype}}) where {stringtype} = stringtype
+isstringtype(::Type{T}) where {stringtype, T<:ListKind{stringtype}} = stringtype
```
quinnj, line 208: This seems like a more complicated equivalent?

quinnj (changes requested):
Was the original issue that we were defining `eltype` on the _instances_ instead of types? That makes sense to me and I agree we should fix. But all the changes like:
```julia
isstringtype(::Type{T}) where {stringtype, T<:ListKind{stringtype}} = stringtype
```
should be reverted, as they're just over-complicating the methods and are functionally equivalent. i.e. in this particular example, `ListKind` is already concrete, so `T <: ListKind` doesn't really match any _additional_ types, so we're just introducing an unnecessary type parameter in `T`.

baumgold:
I reverted the changes you suggested as I can't seem to remember why I suggested them in the first place.

quinnj (approved):

PR #405, opened 2023-03-15
do not instantiate new Vector when calling fromarrow

baumgold:
Currently when there's a table containing a column of type variable-size list, whenever the user accesses an entry in the list a new Vector is instantiated causing allocations and performance degradation.  This PR proposes to prevent instantiating the Vector and instead return to the user a zero-copy view of the data (SubArray) that still abides by the AbstractVector interface.

For example, instead of getting a newly-allocated `Vector{Int64}`:

```julia
julia> using Arrow, Tables

julia> x = (a=[1,2,3], b=[[1,1,1], [2,2], [3,3,3,3]]);

julia> table = Arrow.Table(Arrow.tobuffer(x))
Arrow.Table with 3 rows, 2 columns, and schema:
:a  Int64
:b  Vector{Int64} (alias for Array{Int64, 1})

julia> p = iterate(Tables.rows(table))
(Tables.ColumnsRow{Tables.CopiedColumns{Arrow.Table}}: (a = 1, b = [1, 1, 1]), 2)

julia> p[1].b
3-element Vector{Int64}:
1
1
1
```

we'll instead get a `SubArray`:

```julia
julia> using Arrow, Tables

julia> x = (a=[1,2,3], b=[[1,1,1], [2,2], [3,3,3,3]]);

julia> table = Arrow.Table(Arrow.tobuffer(x))
Arrow.Table with 3 rows, 2 columns, and schema:
:a  Int64
:b  Vector{Int64} (alias for Array{Int64, 1})

julia> p = iterate(Tables.rows(table))
(Tables.ColumnsRow{Tables.CopiedColumns{Arrow.Table}}: (a = 1, b = [1, 1, 1]), 2)

julia> p[1].b
3-element view(::Arrow.Primitive{Int64, Vector{Int64}}, 1:3) with eltype Int64:
1
1
1
```

### DIFF
--- a/src/ArrowTypes/src/ArrowTypes.jl
+++ b/src/ArrowTypes/src/ArrowTypes.jl
@@ -220,8 +220,7 @@ _symbol(ptr, len) = ccall(:jl_symbol_n, Ref{Symbol}, (Ptr{UInt8}, Int), ptr, len
 fromarrow(::Type{Symbol}, ptr::Ptr{UInt8}, len::Int) = _symbol(ptr, len)
 
 ArrowKind(::Type{<:AbstractArray}) = ListKind()
-fromarrow(::Type{A}, x::A) where {A <: AbstractVector{T}} where {T} = x
-fromarrow(::Type{A}, x::AbstractVector{T}) where {A <: AbstractVector{T}} where {T} = convert(A, x)
+fromarrow(::Type{A}, x::AbstractVector{T}) where {T, A <: AbstractVector{T}} = x
 ArrowKind(::Type{<:AbstractSet}) = ListKind()
 ArrowType(::Type{T}) where {T <: AbstractSet{S}} where {S} = Vector{S}
 toarrow(x::AbstractSet) = collect(x)


quinnj:
Sorry for the slow response. Yeah.......I do like this. I think I remember considering doing this too back when I originally implemented this, but I was worried about something, but can't remember now.

I like this change, but I'm worried about it being breaking.

quinnj (commented):
Hmmmm, reviewing this again, I'm not sure `fromarrow` is the right way to change this. It makes me uncomfortable that we wouldn't be following the `fromarrow` "contract" in that it should return an instance of type `A`, but we're saying that as long as it's an `AbstractVector{T}` then that's good enough. Maybe there's something we can do in the list.jl code or eltypes.jl to say that the variable sized list type returns SubArrays?

baumgold:
> Maybe there's something we can do in the list.jl code or eltypes.jl to say that the variable sized list type returns SubArrays?

I spent some time trying to implement that but couldn't find a clean way to do so.  For `juliaeltype` to return SubArray instead of Vector it would need to know the "parent" type of the SubArray ([parent-type is a type-parameter](https://github.com/JuliaLang/julia/blob/v1.9.0/base/subarray.jl#L14-L24)).  However the "parent" type is deeply embedded in return-type of the [build function](https://github.com/apache/arrow-julia/blob/v2.5.2/src/table.jl#L583).  We could refactor the build functions to separate type calculation from type instantiation so that the calculation part could be reused in the `juliaeltype` function, but that seems like a lot of work so I didn't peruse it yet.  Thoughts?

quinnj:
Here's a somewhat simple approach: https://github.com/apache/arrow-julia/pull/446. It's not quite passing things yet; I need to check if the append test error is legit. I think the others are just updates where we're checking the schema strictly.

baumgold:
Closing in place off #446

Issue #376, opened 2023-01-12
Release ArrowTypes@2.0.2


mattBrzezinski:
A place for @kou to post the progress of releasing `ArrowTypes@2.0.2`, based off [this](https://github.com/apache/arrow-julia/commit/94ee070adcf6d4a7ed3108561ee014702e75b835#commitcomment-96003882) comment.

kou:
@quinnj Should we release only `ArrowTypes`? (We don't release `Arrow`?)
Our established release script doesn't assume this case. (`ArrowTypes` is only released.)
Can we also release `Arrow@2.4.2` or something?

quinnj:
Yeah, let's do an Arrow 2.4.2 release as well. https://github.com/apache/arrow-julia/pull/377

kou:
Start voting: https://lists.apache.org/thread/wwctxdr1cqd940z8j7rvfbncsxkyp59h

kou:
Passed: https://lists.apache.org/thread/c4ggkv4vy5kctn2vtgpgg3195t6sn9g1

Published: https://dist.apache.org/repos/dist/release/arrow/arrow-julia-2.4.2/

Could someone update the Julia General registry?

ericphanson:
Registry updated! TagBot is currently malfunctioning so we don’t have a github “release” (#370) but that isn’t required (it’s customary and helps users see the latest release but the package manager doesn’t use it in any way).

kou:
Thanks!

PR #407, opened 2023-03-16
Bump Arrow + ArrowTypes versions

baumgold:


baumgold:
@kou - Can you please help out with a release of ArrowTypes as well?  Thanks.

kou:
Sure! But our release process doesn't support `ArrowTypes` only release yet.
Can we also release `Arrow`?
See also: https://github.com/apache/arrow-julia/issues/376#issuecomment-1381414099

baumgold:
> Sure! But our release process doesn't support `ArrowTypes` only release yet. Can we also release `Arrow`? See also: [#376 (comment)](https://github.com/apache/arrow-julia/issues/376#issuecomment-1381414099)

Sure, that should be okay I think.

baumgold:
@kou - Picking this back up again.  I bumped the version of both Arrow and ArrowTypes.  Is this still the correct thing to do in order to trigger an ArrowTypes release?  If so then I'll merge this PR so you can start a vote.  Thanks.

kou:
Yes!

baumgold:
@kau - I don't yet see the new release of ArrowTypes.  I expect to see v2.1.0 in the [JuliaRegistry](https://github.com/JuliaRegistries/General/blob/master/A/ArrowTypes/Versions.toml), but it doesn't look like it's there yet.

kou:
@ericphanson Could you help us? I think that we may need to change how we use ` @JuliaRegistrator register()`: https://github.com/apache/arrow-julia/commit/686ab570b831035715cb58f666233ec673e50d8f#commitcomment-109739320

baumgold:
@kou - I believe we just have to explicitly register the sub-package by adding the followiong comment [here](https://github.com/apache/arrow-julia/commit/686ab570b831035715cb58f666233ec673e50d8f):

```@JuliaRegistrator register subdir=src/ArrowTypes```

I don't have permissions to do so.  Can you try, please?  Thanks.

kou:
Done: https://github.com/apache/arrow-julia/commit/686ab570b831035715cb58f666233ec673e50d8f#commitcomment-110073911

PR #412, opened 2023-04-03
Add kwarg to filter columns

JoaoAparicio:
Currently we don't have the option to load just a subset of the columns. This matters e.g. when compression is the bottleneck.

For example, create a compressed arrow file.

```julia
using Arrow
p = tempname();
N = 1000000
tbl = (
    a=rand(N),
    b=rand(N),
    c=rand(N),
    d=rand(N),
    e=rand(N),
    f=[rand(rand(0:100)) for _ in 1:N],
);
Arrow.write(p, tbl; compress=:zstd);
```

Column `f` is the longest - it has an expected 50*N elements vs N for the rest Some times we only care for some of the other columns. Currently we must decompress all columns regardless:
```julia
using BenchmarkTools
@btime tbl = Arrow.Table(p);  # 359.205 ms (530 allocations: 794.23 MiB)
```
With this commit we can load only some of the columns
```julia
@btime tbl = Arrow.Table(p; filtercolumns=["a"]);  # 6.146 ms (231 allocations: 14.33 MiB)
```


JoaoAparicio:
https://github.com/apache/arrow-julia/issues/340
https://github.com/apache/arrow-julia/issues/353

JoaoAparicio:
Converting this to draft as I'm working on something that will supersede this.

JoaoAparicio:
Does anyone wanna re-run CI? Looks like macos got stuck

kou:
Done.

PR #419, opened 2023-04-07
fix `Vector{UInt8}` writing

### LINKED ISSUE #411: `Vector{UInt8}` mis-represented when writing to disk
```julia
julia> using Arrow, DataFrames

julia> df = DataFrame(; x = [[0x01, 0x02], UInt8[], [0x03]])
3×1 DataFrame
 Row │ x
     │ Array…
─────┼───────────────────
   1 │ UInt8[0x01, 0x02]
   2 │ UInt8[]
   3 │ UInt8[0x03]

julia> Arrow.write("/tmp/julia.feather", df)
"/tmp/julia.feather"
```

instead of `Vector{UInt8}`, it ended up being seen as `byte-string`
```python
In [1]: import pyarrow.feather

In [3]: pyarrow.feather.read_table("/tmp/julia.feather")["x"]
Out[3]:
<pyarrow.lib.ChunkedArray object at 0x7fb2994c86d0>
[
  [
    0102,
    ,
    03
  ]
]
```

Moelf:
fix #411

Setup:
```julia
julia> using Arrow

julia> df = (; x = [[0x01, 0x02], UInt8[], [0x03]])

julia> Arrow.write("/tmp/julia.feather", df)
"/tmp/julia.feather"
```

Before:
```python

In [3]: pyarrow.feather.read_table("/tmp/julia.feather")["x"]
Out[3]:
<pyarrow.lib.ChunkedArray object at 0x7fb2994c86d0>
[
  [
    0102,
    ,
    03
  ]
]

```

**Note:** `data = (; x = [b"12", b"", b"3"])` from Julia will give write out the exact same thing, which is IMO why this was wrong.

After:
```python
In [39]: pyarrow.feather.read_table("/tmp/julia.feather")["x"]
Out[39]:
<pyarrow.lib.ChunkedArray object at 0x7f4568b376a0>
[
  [
    [
      1,
      2
    ],
    [],
    [
      3
    ]
  ]
]
```


Moelf:
test wouldn't fully pass, could use some help and also this is technically breaking but I believe this is the correct thing to do.

If user wants byte-string, they should use `b""` and not `Vector{UInt8}` @quinnj

src/eltypes.jl
```diff
@@ -393,23 +393,21 @@ end
...
-            end
+    if eltype(A) == UInt8 && T <: AbstractString || T <: Union{AbstractString, Missing}
+        if O == Int32
+            Meta.utf8Start(b)
+            return Meta.Utf8, Meta.utf8End(b), nothing
+        else # if O == Int64
+            Meta.largUtf8Start(b)
+            return Meta.LargeUtf8, Meta.largUtf8End(b), nothing
+        end
+    elseif eltype(A) == UInt8 && ArrowTypes.isstringtype(T)
```
Moelf, line 404: the only additional type cathed here is `Base.CodeUnits`.

Ideally, we want the first (above) used for long string and this one for short-string, but that's not a well defined concept, so for now we just say, "use CodeUnits if you want output byte-string in Arrow"

Moelf:
The reason why this PR doesn't need to touch test is because our implementation was self-consistent in a hard way -- it ignored an entire possible type composition in arrow:
image

Moelf:
@quinnj 👀  gentle bump?

simsurace:
@ericphanson or @baumgold would you be willing/able to review this?

baumgold:
> @ericphanson or @baumgold would you be willing/able to review this?

Looks like some of the tests are failing.  Would you mind investigating and fixing?  Thanks.

Moelf:
@baumgold no they are not

The tests are failing because this changes both code base, you should look at this test:

https://github.com/apache/arrow-julia/actions/runs/4685294460/jobs/8302259625?pr=419

baumgold:
I see.  Probably we should split this up into 2 smaller PRs: one for ArrowTypes and one for Arrow.  Assuming we make a new release of ArrowTypes then in the follow-on Arrow PR we can bump the compat of ArrowTypes to the newly released version.

Moelf:
what?.........

ericphanson:
> isn't the whole point of monorepo so that when we have 2 packages we don't have to make rapid release to fix tests for the other package?

I’m not sure all the reasons for using this setup in this case but that definitely isn’t always a goal. E.g. https://discourse.julialang.org/t/how-beacon-packages-julia-code-in-a-monorepo/90822.

Moelf:
So what people think, I should do a separate PR to add convince function??

simsurace:
If this is what it takes to keep the ball moving, I would suggest doing so, but I am not a maintainer, just an observer interested to see this package continue getting improved.

Moelf:
just to be clear, in that case we will be doing:
1. make a PR to add a convenient function ("will be used in next PR trust me")
2. release a new ArrowTypes.jl for that, and potentially making a bunch of users re-precompile their stuff for no reason
3. see tests in this PR turn all green?

src/eltypes.jl
quinnj: What does "cathed" mean?

Moelf: sorry, "caught", as in caught by the condition

quinnj:
@Moelf, do you mind if I make an alternative PR to fix the original issue here?

Moelf:
go ahead

quinnj:
Ok, PR up: https://github.com/apache/arrow-julia/pull/439. Sorry to be so MIA lately; I've been tied up in some heavy other projects and it's been too hard to context-switch back here. A lot of that work (webstack-related) has wrapped up (mostly), I'm hoping to have more time to help review/fix stuff here.

This PR was definitely in the right direction @Moelf; thanks for the contribution. It was a great starting point. Couple of specific points:
* I'd rather not pun the `ArrowTypes.isstringtype` function for non-ListKind types; it keeps that a little cleaner IMO
* We can make Base.CodeUnits the main translation between the Arrow Binary type
* We can add some compat shims in Arrow to avoid the awkward ArrowTypes/Arrow mis-compat

PR #421, opened 2023-04-09
Add @testsets for misc tests

JoaoAparicio:
Tests under "misc" aren't bundled into testsets. As a consequence, tests terminate when the first misc test fails, making it difficult to get a good picture of how many failing misc tests remain.

In this commit I've bundled misc tests into testsets.

For example, in (unrelated) work that I'm doing I can see what's still missing:
with_testsets

Now I can see I have 5 errors out of 127 total misc tests.
Before it would  just show 1 error out of 69 misc tests, as it was stopping early.
Does anyone else like this?

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -110,7 +110,7 @@ end # @testset "abstract path"
 
 @testset "misc" begin
 
-# multiple record batches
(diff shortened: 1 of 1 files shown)


baumgold (approved):
LGTM

Issue #340, opened 2022-10-06
Feather file with compression and larger than RAM


Moelf:
Last time I checked, `mmap` breaks down for files with compression. This is understandable because the compressed buffers clearly can't be re-interpreted without inflation.

But the larger the file is more likely to be compressed, can we decompress only a single "row group" (and only the relevant columns, of course) on the fly yet? -- this is for the case when a user is doing per-row iteration

if user access the table by `tbl[range, range]`, then clearly, we might need to read more than one row-group and chop off head/tails depending on where the overlap is

quinnj:
Hmmmm......we'll have to see what we can do here. I've had the idea for a while as a Tables.jl-wide feature to support projection/filter push down for sources in a generic way. That would translate really well to Arrow and would allow us to more easily avoid decompressing when not necessary. There's probably more we can do in the short-term though to avoid materializing when not needed.

Moelf:
this is the whole thing we do un UnROOT.jl for a physics-community only thing called `TTree`, their next-gen storage called `RNTuple` is basically `Feather`: https://indico.cern.ch/event/1208767/contributions/5083082/attachments/2523220/4340111/PPP_uproot_RNTuple.pdf#page=13

while we will get there eventually, in `UnROOT` I have the whole machinery basically:
- `getindex` -> find the row group -> 
    - if not in cache, decompress and put in cache
    - if in cache, directly try to locate the slot

this way, at most one row group worth of data ever lives in RAM, in fact that's the minimal amount you need in RAM, because you can only know row number start-end for an entire row group and you have to count inside it.

but yeah, this is a whole thing in UnROOT.jl and it's mission-critical because our data are like, O(100) GB compressed all the time

JoaoAparicio:
> Hmmmm......we'll have to see what we can do here. I've had the idea for a while as a Tables.jl-wide feature to support projection/filter push down for sources in a generic way. That would translate really well to Arrow and would allow us to more easily avoid decompressing when not necessary. There's probably more we can do in the short-term though to avoid materializing when not needed.

Ping for comments on #412.
This isn't the most general filter push down, but it does allow us to avoid unnecessary decompression.

PR #422, opened 2023-04-11
Pre-allocate buffer

JoaoAparicio:
If we let transcode to its own allocation it will allocate a small vector, start filling it, resize the vector, fill it some more, resize the vector, etc.

Instead in this commit we pre-allocate a vector of the corect size and pass it to transcode().

Inspired by https://github.com/apache/arrow-julia/pull/399


JoaoAparicio:
Waiting on release of 
https://github.com/JuliaIO/TranscodingStreams.jl/pull/136

Moelf:
any thoughts on:
- https://github.com/apache/arrow-julia/issues/340

?

JoaoAparicio:
I have some thoughts. One solution to "my dataset is larger than memory" is partitioning. If your dataset is partitioned in such a way that each partition fits in memory, you can iterate it with
```
stream = Arrow.Stream(path)
for tbl in stream
    ...
end
```
You can do this right now without requiring any additional features from this package.

In contrast what is discussed in #340 (which is: don't decompress if you don't have to) is a different approach, but doesn't yet exist.

Currently I have some commits that add the feature to multi-thread decompression at the buffer level. I will be trying to upstream what I have so far. The difficulty is that these commits touch a lot of code, so this won't happen overnight. I imagine couple of weeks? On top of that it should be straightforward to implement what is discussed in #340.

Moelf:
```
stream = Arrow.Stream(path)
for tbl in stream
```

the problem of this approach is it's decompressing every branch. Consider examples such as:
- https://github.com/apache/arrow-julia/issues/417

decompressing every branch would be super slower

ericphanson:
nope, LGTM!

PR #424, opened 2023-04-11
add dev CI for monorepo

Moelf:
@ericphanson

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,7 @@ jobs:
     name: Monorepo dev - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         version:
           - '1' # automatically expands to the latest stable 1.x release of Julia
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,43 @@ jobs:
       - uses: codecov/codecov-action@v1
         with:
           file: lcov.info
+  test_monorepo:
+    name: Monorepo dev - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1' # automatically expands to the latest stable 1.x release of Julia
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: actions/cache@v1
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - name: Dev monorepo dependencies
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -105,6 +105,43 @@ jobs:
...
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - name: Dev monorepo dependencies
+        shell: julia --project=monorepo {0}
+        run: |
+          using Pkg;
+          pkg"dev . ./ArrowTypes"
```
ericphanson, line 140: better to use the API, and also the path is in `src`:
```suggestion
          Pkg.develop([PackageSpec(path="."), PackageSpec(path="src/ArrowTypes")])
```

Moelf:
https://github.com/apache/arrow-julia/actions/runs/4669945251/jobs/8269066505#step:6:166

wait it errored but didn't fail CI? and also, why would this error??

Moelf:
this is checking out the source code for both packages

Moelf:
ok, it's because the test is directly including `test.jl` from ArrowTypes, which bypassed the `using Sockets` in `runtests.jl`

added now directly in Arrow.jl's `runtests.jl` as we do for UUID already

tbh I don't understand why the tests are so messy here

Moelf:
that is fixed now with `fast-fail`

ericphanson:
I think maybe this should be an entirely separate workflow file. Because the situation in #419 is that it works when you use them together, but fails when used release-vs-dev versions, because you're updating them in tandem. So with `fast-fail:true` we will only see one of those things, not both. And it seems `fast-fail: false` is problematic bc it shows stuff as passing when it isn't.

Moelf:
the `fast-fail` only apply to `ci - monorepo`, it does not affect other step like `test`

Moelf:
ah, it's the `continue-on-error`

Moelf:
finally!

ericphanson:
np, thanks for your persistence. Can we verify the failure mode here? E.g. put an `error()` in an ArrowTypes function (or even `__init__() = error()` in ArrowTypes), so we can see the expected workflows pass/fail. In that case we should expect: Arrow.jl tests to pass (uses release ArrowTypes module), ArrowTypes to fail (uses dev version) and monorepo tests to fail (uses dev version). We can verify we get the appropriate ❌ here.

Moelf:
this should work

ericphanson:
Great, I see ArrowTypes see failing, monorepo CI failing, release CI failing, but Arrow.jl CI passing, as expected. Ok go ahead and remove the test error and we can merge IMO

ericphanson (approved):

.github/workflows/ci.yml
```diff
@@ -105,6 +105,43 @@ jobs:
...
+      fail-fast: true
+      matrix:
+        version:
+          - '1' # automatically expands to the latest stable 1.x release of Julia
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@v2
```
kou, line 121:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/ci.yml
```diff
@@ -105,6 +105,43 @@ jobs:
...
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: actions/cache@v1
```
kou, line 126:
```suggestion
      - uses: actions/cache@v3
```

kou (commented):
How about enabling Dependabot for GitHub Actions?

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#example-dependabotyml-file-for-github-actions

Moelf:
can we merge this?

kou:
Ah, sorry. I should have said explicitly that Dependabot should be done as a follow-up PR not in this PR.
Creating a new issue for it is enough for now.

(If we don't need Dependabot, we don't need to do anything.)

PR #428, opened 2023-04-14
Base.isdone for Stream

baumgold:
Implement Base.isdone for Arrow.Stream

https://github.com/JuliaLang/julia/blob/v1.8.5/base/essentials.jl#L882-L895

This enables calling `Base.isempty(::Arrow.Stream)` without affecting the state of the Arrow.Stream object.

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -87,9 +87,17 @@ function Stream(inputs::Vector{ArrowBlob}; convert::Bool=true)
     Stream(inputs, inputindex, batchiterator, names, types, schema, dictencodings, dictencoded, convert, compression)
 end
 
-Stream(input, pos::Integer=1, len=nothing; kw...) = Stream([ArrowBlob(tobytes(input), pos, len)]; kw...)
-Stream(input::Vector{UInt8}, pos::Integer=1, len=nothing; kw...) = Stream([ArrowBlob(tobytes(input), pos, len)]; kw...)
-Stream(inputs::Vector; kw...) = Stream([ArrowBlob(tobytes(x), 1, nothing) for x in inputs]; kw...)
+function Stream(input, pos::Integer=1, len=nothing; kw...)
+    b = tobytes(input)
+    isempty(b) ? Stream(ArrowBlob[]; kw...) : Stream([ArrowBlob(b, pos, len)]; kw...)
+end
+
+function Stream(input::Vector{UInt8}, pos::Integer=1, len=nothing; kw...)
+    b = tobytes(input)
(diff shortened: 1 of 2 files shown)


src/table.jl
```diff
@@ -80,9 +80,17 @@ function Stream(inputs::Vector{ArrowBlob}; convert::Bool=true)
...
+    b = tobytes(input)
+    isempty(b) ? Stream(ArrowBlob[]; kw...) : Stream([ArrowBlob(b, pos, len)]; kw...)
+end
+
+function Stream(input::Vector{UInt8}, pos::Integer=1, len=nothing; kw...)
+    b = tobytes(input)
+    isempty(b) ? Stream(ArrowBlob[]; kw...) : Stream([ArrowBlob(b, pos, len)]; kw...)
+end
+
+Stream(inputs::AbstractVector; kw...) = Stream([ArrowBlob(tobytes(x), 1, nothing) for x in inputs]; kw...)
```
quinnj, line 93: Do we not want to do a similar check for each input element?

quinnj:
Hmmmm, I'm not sure I'm following the changes or reasoning here. Can you explain how `isempty` changes the current `Arrow.Stream` state? I'm not super familiar w/ `Base.isdone` or how it's used for other types or why; if you could share an example, that would be great. Another question that comes to mind is why we shouldn't define `isempty(::Arrow.Stream)` ourselves if it's causing problems?

baumgold:
@quinnj - Iterating on Arrow.Stream modifies the `inputindex` field, making it stateful iteration rather than stateless.  The default implementation for Base.isempty uses iteration to check if there are further elements remaining:

https://github.com/JuliaLang/julia/blob/v1.8.5/base/essentials.jl#L785-L789

Unfortunately this doesn't work for stateful iterators since calling Base.isdone modifies the iterator's state.  A solution to this is to define Base.isdone:

https://github.com/JuliaLang/julia/blob/v1.8.5/base/essentials.jl#L882-L895

Note the doc-string explains this quite clearly:

> This function provides a fast-path hint for iterator completion. This is useful for mutable iterators that want to avoid having elements consumed, if they are not going to be exposed to the user (e.g. to check for done-ness in `isempty` or `zip`). Mutable iterators that want to opt into this feature should define an isdone method that returns true/false depending on whether the iterator is done or not. Stateless iterators need not implement this function. If the result is `missing`, callers may go ahead and compute `iterate(x, state...) === nothing` to compute a definite answer.

src/table.jl
quinnj: Bump; it seems like we should filter out empty elements in the `inputs` array here?

baumgold: Agreed, fixed.  Thanks.

quinnj (approved):

Issue #420, opened 2023-04-09
Test failure when running with > 1 thread


JoaoAparicio:
Found a flaky test, this one
https://github.com/apache/arrow-julia/blob/c469151d4ff261b50c59bf98101f068fa577fca4/test/runtests.jl#L297-L305

If you run it without threads, always passes. If you run it with `-t 10` it has a significant chance of erroring.

A) Start Julia with `-t 10`

B) Create the environment with:
```julia-repl
]activate --temp
add Arrow
add Tables
add PooledArrays
```

C) Run the test many times:
```julia
using Tables
using Arrow
using PooledArrays

for _ in 1:10000
    t = Tables.partitioner(
        (
            (a=Arrow.toarrowvector(PooledArray([1,2,3  ])),),
            (a=Arrow.toarrowvector(PooledArray([1,2,3,4])),),
            (a=Arrow.toarrowvector(PooledArray([1,2,3,4,5])),),
        )
    )
    tt = Arrow.Table(Arrow.tobuffer(t))
end
```

Highly likely that you'll hit this:

```julia
(jl_k3aixG) pkg> ERROR: TaskFailedException
Stacktrace:
 [1] wait
   @ ./task.jl:349 [inlined]
 [2] Arrow.Table(blobs::Vector{Arrow.ArrowBlob}; convert::Bool)
   @ Arrow ~/.julia/packages/Arrow/P0wVk/src/table.jl:386
 [3] Table
   @ ~/.julia/packages/Arrow/P0wVk/src/table.jl:295 [inlined]
 [4] Arrow.Table(input::IOBuffer, pos::Int64, len::Nothing; kw::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Arrow ~/.julia/packages/Arrow/P0wVk/src/table.jl:290
 [5] Table
   @ ~/.julia/packages/Arrow/P0wVk/src/table.jl:290 [inlined]
 [6] Arrow.Table(input::IOBuffer)
   @ Arrow ~/.julia/packages/Arrow/P0wVk/src/table.jl:290
 [7] top-level scope
   @ ./Untitled-2:13

    nested task error: MethodError: no method matching resize!(::Arrow.DictEncoded{Int64, Int8, Arrow.Primitive{Int64, Vector{Int64}}}, ::Int64)
    
    Closest candidates are:
      resize!(::Vector, ::Integer)
       @ Base array.jl:1246
[...]

PR #431, opened 2023-04-23
Run with 1 and 2 threads during tests

### LINKED ISSUE #426: CI doesn't test with multiple threads
which leads to issues like https://github.com/apache/arrow-julia/issues/420. We should run our tests with both one and two threads (since that's what GitHub supports, IIUC).

quinnj:
Fixes #426


ericphanson (approved):
lgtm, and I can see it passed through: https://github.com/apache/arrow-julia/actions/runs/4776416247/jobs/8491432668?pr=431#step:6:33

it's too bad this isn't enough to reproduce #420

baumgold (approved):

PR #436, opened 2023-05-04
Add handling of len = -1 in uncompress

### LINKED ISSUE #435: Unhandled sentinel value for len in compression causes invalid Array dimensions
I'm generating a bunch of Arrow files from the apache java implementation and many of them are not readable by Arrow.jl (but they are readable by the java implementation).

When following the java decoding process in the debugger, it seems that both implementations agree up to the following line in the java implementation:
https://github.com/apache/arrow/blob/febd0ff144cfb8b2baffb1cb0be57ca40dc7cc77/java/vector/src/main/java/org/apache/arrow/vector/compression/AbstractCompressionCodec.java#L72-L75

It seems like length == -1 is some kind of sentinel value for no compression (maybe the compressor gave up or something?) which does not seem to be handled in the corresponding function in Arrow.jl: 
https://github.com/apache/arrow-julia/blob/e893c327f177f5a4d5efeab831df0fe93ab4ec5b/src/table.jl#L521-L524

I have verified that Arrow.jl indeed does read out len = -1 (which in turn causes an error saying `invalid Array dimensions` when creating the decodedbytes vector).

DrChainsaw:
Fix #435 

Hipshot PR from the github API. Not sure how to add tests for this if needed. It reads my files correctly at least :)

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -523,7 +523,7 @@ function uncompress(ptr::Ptr{UInt8}, buffer, compression)
     encodedbytes = unsafe_wrap(Array, ptr, buffer.length - 8)
     if len === -1
         # len = -1 means data is not compressed
-        return length(encodedbytes), encodedbytes
+        return length(encodedbytes), copy(encodedbytes)
     end
                                         
     decodedbytes = Vector{UInt8}(undef, len)
--- a/src/table.jl
+++ b/src/table.jl
@@ -521,6 +521,11 @@ function uncompress(ptr::Ptr{UInt8}, buffer, compression)
     len = unsafe_load(convert(Ptr{Int64}, ptr))
     ptr += 8 # skip past uncompressed length as Int64
     encodedbytes = unsafe_wrap(Array, ptr, buffer.length - 8)
+    if len === -1
+        # len = -1 means data is not compressed
+        return length(encodedbytes), encodedbytes
+    end
+                                        
     decodedbytes = Vector{UInt8}(undef, len)
     if compression.codec === Meta.CompressionTypes.LZ4_FRAME
         transcode(LZ4FrameDecompressor, encodedbytes, decodedbytes)


src/table.jl
```diff
@@ -521,6 +521,11 @@ function uncompress(ptr::Ptr{UInt8}, buffer, compression)
     len = unsafe_load(convert(Ptr{Int64}, ptr))
     ptr += 8 # skip past uncompressed length as Int64
     encodedbytes = unsafe_wrap(Array, ptr, buffer.length - 8)
+    if len === -1
```
baumgold, line 524: How do we end up in the `uncompress` method for date that is not compressed?  I believe we should be checking for compression earlier than here.  Is that a branch where this check is missing?

DrChainsaw: I don't know the logic why the writer would do this. I encountered the issue when trying to read files generated from the official java arrow package. When stepping through the debugger during reading in java I encountered this lines which I just copied over to the Julia implementation: https://github.com/apache/arrow/blob/febd0ff144cfb8b2baffb1cb0be57ca40dc7cc77/java/vector/src/main/java/org/apache/arrow/vector/compression/AbstractCompressionCodec.java#L72-L75

In the linked issue there is an example file which also reads in pyarrow. I haven't made any effort to find the corresponding line in there though.

DrChainsaw: Comment about the flag which might shed some light: https://github.com/apache/arrow/blob/cd6e2a4d2b9373b942da18b4cc82cb41431764d9/java/vector/src/main/java/org/apache/arrow/vector/compression/CompressionUtil.java#L67-L71

It could be that the java implementation just does its own thing here without regards to any standard, wouldn't that be a pretty irresponsible thing to do given that the repo markets itself as being cross platform.

DrChainsaw:
CI timeout on macos seemed to happen during precompilation of dependencies. Trying to restart through open-close.

quinnj (approved):
Super weird that Java does this but oh well, it's easy enough to handle on our end. Thanks for the fix here @DrChainsaw; I'll add a test for this in a follow up PR.

PR #438, opened 2023-05-18
Update ArrowTypes.jl to add utility function

Moelf:
split from #419


quinnj:
Not needed wtih https://github.com/apache/arrow-julia/pull/439

PR #439, opened 2023-05-18
Don't treat Vector{UInt8} as Arrow Binary type

### LINKED ISSUE #411: `Vector{UInt8}` mis-represented when writing to disk

quinnj:
Fixes #411. Alternative to #419.

This PR should be compatible with or without the ArrowTypes changes. I think it's fine to do compat things in Arrow like this as long as they don't get out of hand and we can eventually remove them as we bump required ArrowTypes versions and such.

The PR consists of not treating `Vector{UInt8}` as the Arrow Binary type, which is meant for "binary string"s. Julia has a pretty good match for that in `Base.CodeUnits`, so instead, we use that to write Binary and `Vector{UInt8}` is treated as a regular List of Primitive UInt8 type.

### DIFF
--- a/src/ArrowTypes/src/ArrowTypes.jl
+++ b/src/ArrowTypes/src/ArrowTypes.jl
@@ -208,8 +208,11 @@ isstringtype(::ListKind{stringtype}) where {stringtype} = stringtype
 isstringtype(::Type{ListKind{stringtype}}) where {stringtype} = stringtype
 
 ArrowKind(::Type{<:AbstractString}) = ListKind{true}()
+# Treate Base.CodeUnits as Binary arrow type
+ArrowKind(::Type{<:Base.CodeUnits}) = ListKind{true}()
 
 fromarrow(::Type{T}, ptr::Ptr{UInt8}, len::Int) where {T} = fromarrow(T, unsafe_string(ptr, len))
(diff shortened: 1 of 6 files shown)


quinnj:
Shoot, I forgot we run tests both ways; I'll add some more compat here

quinnj:
This should be ready to review/merge; cc: @Moelf @baumgold

Moelf (approved):

PR #443, opened 2023-05-24
Add Tables.partitions definition for Arrow.Table

### LINKED ISSUE #293: Support Tables.partitions when reading "arrow file" format in addition to "stream" format

quinnj:
We had this functionality w/ `Arrow.Stream`, but it's convenient and not that expensive to define it for `Arrow.Table` as well.

Fixes #293.

### DIFF
--- a/src/table.jl
+++ b/src/table.jl
@@ -261,6 +261,7 @@ types(t::Table) = getfield(t, :types)
 columns(t::Table) = getfield(t, :columns)
 lookup(t::Table) = getfield(t, :lookup)
 schema(t::Table) = getfield(t, :schema)
+metadata(t::Table) = getfield(t, :metadata)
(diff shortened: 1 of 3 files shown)


quinnj:
Nightly failures are https://github.com/JuliaLang/julia/pull/49777, which should be resolved soon

PR #446, opened 2023-06-01
Return SubArrays when possible for arrow list types

quinnj:


src/append.jl
```diff
@@ -197,7 +197,13 @@ end
 function is_equivalent_schema(sch1::Tables.Schema, sch2::Tables.Schema)
     (sch1.names == sch2.names) || (return false)
     for (t1,t2) in zip(sch1.types, sch2.types)
-        (t1 === t2) || (return false)
+        tt1 = Base.nonmissingtype(t1)
+        tt2 = Base.nonmissingtype(t2)
+        if t1 == t2 || (tt1 <: AbstractVector && tt2 <: AbstractVector && eltype(tt1) == eltype(tt2))
```
quinnj, line 202: This is needed to loosen what we consider "equivalent schemas"; i.e. a column w/ eltype `Vector{Int}` is now "equal" to `SubArray{Int, 1, ...}`

quinnj:
@baumgold, this is ready to review if you have a moment.

src/append.jl
baumgold, line 202: This looks good to me.  Small nit-pick... small simplification without the continue:

```julia
t1 != t2 && (tt1 <: AbstractVector && tt2 <: AbstractVector && eltype(tt1) != eltype(tt2)) && return false
```

src/table.jl
```diff
@@ -637,16 +639,23 @@ function build(f::Meta.Field, L::ListTypes, batch, rb, de, nodeidx, bufferidx, c
     bufferidx += 1
     len = rb.nodes[nodeidx].length
     nodeidx += 1
+    meta = buildmetadata(f.custom_metadata)
     if L isa Meta.Utf8 || L isa Meta.LargeUtf8 || L isa Meta.Binary || L isa Meta.LargeBinary
         buffer = rb.buffers[bufferidx]
         bytes, A = reinterp(UInt8, batch, buffer, rb.compression)
         bufferidx += 1
+        T = juliaeltype(f, meta, convert)
```
baumgold, line 647: This can move out of the if/else statement to reduce duplication since it's used in both the if and the else.

baumgold (approved):
This looks good to me, just some small comments.  Thank you for this - it's much cleaner and simpler than what I was working on in #405.

Issue #458, opened 2023-06-07
Error with v2.6.0


baumgold:
After upgrading to Arrow v2.6.0 I get the following error in one of my apps:

```julia
[...]

PR #459, opened 2023-06-08
Fix case of recursively nested lists

quinnj:
Fixes issue reported by @Moelf in #458.

The issue is we need to get the eltype from the recursively built child array instead of as a top-level translation of the current field type.


Moelf:
thx for the swift fix

quinnj:
> what happened to coverage 😆

I think it's getting confused w/ ArrowTypes.jl CI jobs finishing first and then we're getting 0 coverage on _Arrow.jl_ files from ArrowTypes tests.

PR #464, opened 2023-06-12
Formatting

### LINKED ISSUE #398: Introduce automatic code formatting with JuliaFormatter.toml
Would it be possible to pick a code style for automatic formatting with [JuliaFormatter](https://domluna.github.io/JuliaFormatter.jl/dev/) and codify this choice in a `.JuliaFormatter.toml` file?

I believe it would improve the consistency of the code style, but, most importantly, make it easier for contributors to adhere to the same style (and, hence, improve their quality and cadence).

For example, whenever I want to contribute / experiment with changes, I have to disable my auto-formatting (as there is no .JuliaFormatter.toml file) and make all formatting changes manually. I'm certain people will often miss some style issues, so it will lead to repeated PR edits just to align the style.

baumgold:
I used the default JuliaFormatter options, but we can customize them if we want to with a `.JuliaFormatter.toml` file.  Details available here:

https://domluna.github.io/JuliaFormatter.jl

Fixes #398.  CC: @svilupp

### DIFF
--- a/src/FlatBuffers/FlatBuffers.jl
+++ b/src/FlatBuffers/FlatBuffers.jl
@@ -24,7 +24,7 @@ const VOffsetT = UInt16
 const VtableMetadataFields = 2
 
 basetype(::Enum) = UInt8
-basetype(::Type{T}) where {T <: Enum{S}} where {S} = S
+basetype(::Type{T}) where {T<:Enum{S}} where {S} = S
 
 function readbuffer(t::AbstractVector{UInt8}, pos::Integer, ::Type{Bool})
     @inbounds b = t[pos + 1]
@@ -46,7 +46,12 @@ function Base.show(io::IO, x::TableOrStruct)
     if isempty(propertynames(x))
         print(io, "()")
     else
-        show(io, NamedTuple{propertynames(x)}(Tuple(getproperty(x, y) for y in propertynames(x))))
+        show(
+            io,
+            NamedTuple{propertynames(x)}(
+                Tuple(getproperty(x, y) for y in propertynames(x)),
+            ),
+        )
     end
 end
 
--- a/src/FlatBuffers/builder.jl
+++ b/src/FlatBuffers/builder.jl
@@ -20,10 +20,8 @@ const fileIdentifierLength = 4
 Scalar
 A Union of the Julia types `T <: Number` that are allowed in FlatBuffers schema
 """
-const Scalar = Union{Bool,
-Int8, Int16, Int32, Int64,
-UInt8, UInt16, UInt32, UInt64,
-Float32, Float64, Enum}
+const Scalar =
+    Union{Bool,Int8,Int16,Int32,Int64,UInt8,UInt16,UInt32,UInt64,Float32,Float64,Enum}
 
 """
 Builder is a state machine for creating FlatBuffer objects.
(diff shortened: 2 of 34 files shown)


src/FlatBuffers/builder.jl
```diff
@@ -63,8 +71,8 @@ end
 Base.write(sink::Builder, o, x::Union{Bool,UInt8}) = sink.bytes[o+1] = UInt8(x)
 function Base.write(sink::Builder, off, x::T) where {T}
     off += 1
-    for (i, ind) = enumerate(off:(off + sizeof(T) - 1))
-        sink.bytes[ind] = (x >> ((i-1) * 8)) % UInt8
+    for (i, ind) in enumerate(off:(off+sizeof(T)-1))
+        sink.bytes[ind] = (x >> ((i - 1) * 8)) % UInt8
```
quinnj, line 75: What the? what's the rule here? it seems to usually take away spaces in the 1st line, but _adds_ them in the 2nd?

src/FlatBuffers/table.jl
```diff
@@ -72,7 +76,7 @@ function bytevector(t::Table, off)
     off += get(t, off, UOffsetT)
     start = off + sizeof(UOffsetT)
     len = get(t, off, UOffsetT)
-    return view(bytes(t), (start + 1):(start + len + 1))
+    return view(bytes(t), (start+1):(start+len+1))
```
quinnj, line 79: Maybe it's range-specific where it prefers to remove spaces?

quinnj:
I think this is mostly great and I agree we should do it. I think the only change I'm really against is removing spaces around operators, regardless of where or what kind of clause they occur in. I also personally prefer keeping spaces around the `<:` operator in type parameter clauses, but I don't feel as strongly about that if people prefer otherwise.

quinnj:
@baumgold, can you clarify the CI change here? Does it automatically add a commit w/ formatting changes to your PR? Or post suggestions that can be committed?

src/FlatBuffers/FlatBuffers.jl
```diff
@@ -24,10 +24,10 @@ const VOffsetT = UInt16
 const VtableMetadataFields = 2
 
 basetype(::Enum) = UInt8
-basetype(::Type{T}) where {T <: Enum{S}} where {S} = S
+basetype(::Type{T}) where {T<:Enum{S}} where {S} = S
 
 function readbuffer(t::AbstractVector{UInt8}, pos::Integer, ::Type{Bool})
-    @inbounds b = t[pos + 1]
+    @inbounds b = t[pos+1]
```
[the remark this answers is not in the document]
baumgold: I added a `.JuliaFormatter.toml` file and set `whitespace_ops_in_indices = true` to accomplish this

baumgold:
> I think this is mostly great and I agree we should do it. I think the only change I'm really against is removing spaces around operators, regardless of where or what kind of clause they occur in. I also personally prefer keeping spaces around the `<:` operator in type parameter clauses, but I don't feel as strongly about that if people prefer otherwise.

I believe the only control for this that we are given in JuliaFormatter is the following:

https://github.com/domluna/JuliaFormatter.jl#whitespace_ops_in_indices

I overrode the default from false to true.  That seems like it should mostly fix this issue.

baumgold:
> @baumgold, can you clarify the CI change here? Does it automatically add a commit w/ formatting changes to your PR? Or post suggestions that can be committed?

I think the CI change can be simplified using the "julia-format" GitHub action.  I'll try to get this setup shortly.

https://github.com/julia-actions/julia-format

baumgold:
I was trying to using the `julia-format` GitHub action but I can't seem to get it to work.  Instead I took inspiration from the following example but simplified it as much as possible for use here.

https://github.com/julia-actions/julia-format/blob/master/workflows/format_check.yml

ericphanson:
Something I often do at work is check in a `format` directory with a Project and Manifest, a small `run.jl` script that does the formatting and a readme to explain how. Then the CI job can use the same project/manifest to do the check. This can allow folks to easily run the formatting locally and also ensure that JuliaFormatter updates don’t suddenly cause the CI check to fail on unrelated PRs. (Then if eg a new contributor’s PR fails the format check there is an easy remedy for them - to run the local run script in the format project).

One needs to manually PR the format manifest to update it for newer JuliaFormatter versions but imo that’s worth it being explicit (and you only need to update if there’s some specific change you want).

baumgold:
@ericphanson - That script you speak of is embedded in the ci.yml file:

https://github.com/apache/arrow-julia/pull/464/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR165-R179

IMHO it's quite a trivial script.  We can explicitly pin to a version of JuliaFormatter if we want but that adds extra maintenance overhead so would prefer not to unless there's a good reason.  What do you think?  Is this sufficient?

quinnj:
Go for it

PR #477, opened 2023-07-27
Drop missing from `juliaeltype` if `ValidityBitmap` is all valid.

### LINKED ISSUE #373: Addtional `Missing` gets injected into Schema
```julia
julia> rnt.vector_variant_int64_string
5-element RNTupleField{Vector{Union{Int64, String}}}:
 Union{Int64, String}["one"]
 Union{Int64, String}["one", 2]
 Union{Int64, String}["one", 2, 3]
 Union{Int64, String}["one", 2, 3, 4]
 Union{Int64, String}["one", 2, 3, 4, 5]

julia> DataFrame(rnt).vector_variant_int64_string
5-element Vector{Vector{Union{Int64, String}}}:
 ["one"]
 ["one", 2]
 ["one", 2, 3]
 ["one", 2, 3, 4]
 ["one", 2, 3, 4, 5]

julia> Arrow.write(a, DataFrame(rnt))
"/tmp/jl_Lk5W1G92XO"

julia> Arrow.Table(a)
Arrow.Table with 5 rows, 13 columns, and schema:
 :string                       String
 :vector_int32                 Vector{Int32} (alias for Array{Int32, 1})
 :array_float                  Vector{Float32} (alias for Array{Float32, 1})
 :vector_vector_int32          Vector{Vector{Int32}} (alias for Array{Array{Int32, 1}, 1})
 :vector_string                Vector{String} (alias for Array{String, 1})
 :vector_vector_string         Vector{Vector{String}} (alias for Array{Array{String, 1}, 1})
 :variant_int32_string         Union{Missing, Int32, String}
 :vector_variant_int64_string  Vector{Union{Missing, Int64, String}} (alias for Array{Union{Missing, Int64, String}, 1})
 :tuple_int32_string           NamedTuple{(:_0, :_1), Tuple{Int32, String}}
 :pair_int32_string            NamedTuple{(:_0, :_1), Tuple{Int32, String}}
 :vector_tuple_int32_string    Vector{NamedTuple{(:_0, :_1), Tuple{Int32, String}}} (alias for Array{NamedTuple{(:_0, :_1), Tuple{Int32, String}}, 1})
 :lorentz_vector               NamedTuple{(:pt, :eta, :phi, :mass), NTuple{4, Float32}}
 :array_lv                     Vector{NamedTuple{(:pt, :eta, :phi, :mass), NTuple{4, Float32}}} (alias for Array{NamedTuple{(:pt, :eta, :phi, :mass), NTuple{4, Float32}}, 1})
```

### LINKED ISSUE #384: Nullable fields don't always need Union{Missing, T}
I'm trying to implement the GeoArrow spec, which gives back coordinates in a deeply nested list of a FixedList (a point). Because these lists are theoretically nullable, in Julia we get an deeply nested list with `Union`s of `Missing`, even though these vectors contain no `missing`s. An example for a column of LineStrings (there are geometry types that require two more levels of nesting):

```2-element Arrow.List{Vector{Union{Missing, Vector{Union{Missing, Tuple{Float64, Float64}}}}}```

It's pretty hard to convert these elements to a concrete `Vector{Vector{NTuple, Float64}}` without allocating. Is there a way to edit the view to be non missing? An alternative way would be to pass   `all(validitybitmap)` in `build` to `juliaeltype`, so we only set Missing when there are actual missing values.

I'm happy to make a PR if there's consensus on what to do.

Might be related to #373.

### PR DESCRIPTION
**Pull request by evetion**
Fixes #384 
Fixes #373

This does fail a number of tests, which we can fix after we agree on the direction of this PR.

### DIFF
--- a/src/arraytypes/arraytypes.jl
+++ b/src/arraytypes/arraytypes.jl
@@ -200,6 +200,7 @@ end
 
 Base.size(p::ValidityBitmap) = (p.ℓ,)
 nullcount(x::ValidityBitmap) = x.nc
+Base.all(x::ValidityBitmap) = x.nc == 0
 
 function ValidityBitmap(x)
     T = eltype(x)
--- a/src/eltypes.jl
+++ b/src/eltypes.jl
@@ -34,13 +34,18 @@ arrowtype(b, col::AbstractVector{T}) where {T} = arrowtype(b, maybemissing(T))
 arrowtype(b, col::DictEncoded) = arrowtype(b, col.encoding.data)
 arrowtype(b, col::Compressed) = arrowtype(b, col.data)
 
-function juliaeltype(f::Meta.Field, ::Nothing, convert::Bool)
-    T = juliaeltype(f, convert)
+function juliaeltype(f::Meta.Field, ::Nothing, convert::Bool, allvalid)
+    T = juliaeltype(f, convert, allvalid)
     return convert ? finaljuliatype(T) : T
 end
 
-function juliaeltype(f::Meta.Field, meta::AbstractDict{String,String}, convert::Bool)
-    TT = juliaeltype(f, convert)
+function juliaeltype(
+    f::Meta.Field,
+    meta::AbstractDict{String,String},
+    convert::Bool,
+    allvalid::Bool=false,
+)
+    TT = juliaeltype(f, convert, allvalid)
     !convert && return TT
     T = finaljuliatype(TT)
     if haskey(meta, "ARROW:extension:name")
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by evetion** [comment]
Fixes #283

File: src/arraytypes/arraytypes.jl [hunk 1]
```diff
@@ -200,6 +200,7 @@ end
 
 Base.size(p::ValidityBitmap) = (p.ℓ,)
 nullcount(x::ValidityBitmap) = x.nc
+Base.all(x::ValidityBitmap) = x.nc == 0
```
quinnj, line 203: I don't love overloading `all` here; can we just call this `allvalid`? I don't think we need to overload anything for this.

src/eltypes.jl
```diff
@@ -540,13 +555,13 @@ function UnionT(f::Meta.Field, convert)
     UT = UnionT{
         f.type.mode,
         typeids,
-        Tuple{(juliaeltype(x, buildmetadata(x), convert) for x in f.children)...},
+        Tuple{(juliaeltype(x, buildmetadata(x), convert, false) for x in f.children)...},
```
quinnj, line 558: Can you help me understand why we hard-code `false` in a number of places for the `allvalid` argument? vs. in other places calling `all(validity)`?

quinnj (commented):
I think I'm fine going this direction. I'd like to think a bit on if there may be unintended consequences here, but it seems like it should be strictly an improvement.

evetion:
Ok, I will fix the tests 😃

PR #479, opened 2023-07-29
Fix documentation example for Arrow.Writer

guilhermebodin:


baumgold (approved):

guilhermebodin:
Hi @baumgold this is actually inaccurate, the real example should be
```julia
open(Arrow.Writer, tempname()) do writer
    partition1 = (col1 = [1, 2], col2 = ["A", "B"])
    Arrow.write(writer, partition1)
    partition2 = (col1 = [3, 4], col2 = ["C", "D"])
    Arrow.write(writer, partition2)
end
```

PR #487, opened 2023-09-15
Docs: add a note about using `copy()` to get a `DataFrame` where the columns are regular `Vector`s

DilumAluthge:


DilumAluthge:
cc: @quinnj @bkamins

docs/src/manual.md
```diff
@@ -66,6 +66,7 @@ So, what can you do with an `Arrow.Table` full of data? Quite a bit actually!
 Because `Arrow.Table` implements the [Tables.jl](https://juliadata.github.io/Tables.jl/stable/) interface, it opens up a world of integrations for using arrow data. A few examples include:
 
 * `df = DataFrame(Arrow.Table(file))`: Build a [`DataFrame`](https://juliadata.github.io/DataFrames.jl/stable/), using the arrow vectors themselves; this allows utilizing a host of DataFrames.jl functionality directly on arrow data; grouping, joining, selecting, etc.
+* `df = copy(DataFrame(Arrow.Table(file)))`: Build a [`DataFrame`](https://juliadata.github.io/DataFrames.jl/stable/), where the columns are regular `Vector`s. This requires that you have enough memory to load the entire DataFrame into memory.
```
DilumAluthge, line 69: @bkamins Should this say `copy(DataFrame(...); copycols=true)` instead, just to be safe?

Is there any chance that the default value (`copycols=true`) might change to `copycols=false` in the future?

bkamins: This is OK as is. The only thing that @quinnj should confirm is that Arrow columns when copied always produce `Vector` (as maybe they could produce also some other vector type e.g. `PooledVector`?)

quinnj: @bkamins is correct that DictEncoded arrow arrays are materialized as `PooledVector`s.

DilumAluthge: Hmmm. Should we instead say "where the columns are regular `Vector`s and/or `PooledVector`s"?

Or should we just say something like "where the columns are in-memory vectors (instead of the arrow vectors)"?

bkamins: I thought of it and it seems that these are only going to be `Vector`s or `PooledVector`s, and it is not going to change in the future. Right @quinnj? In this case I think it is OK to be specific (the first option)

quinnj: Yeah, that sounds good to me

DilumAluthge: @quinnj @bkamins `PooledVector` comes from PooledArrays.jl (https://github.com/JuliaData/PooledArrays.jl), right?

bkamins: right

docs/src/manual.md
```diff
@@ -66,6 +66,7 @@ So, what can you do with an `Arrow.Table` full of data? Quite a bit actually!
 Because `Arrow.Table` implements the [Tables.jl](https://juliadata.github.io/Tables.jl/stable/) interface, it opens up a world of integrations for using arrow data. A few examples include:
 
 * `df = DataFrame(Arrow.Table(file))`: Build a [`DataFrame`](https://juliadata.github.io/DataFrames.jl/stable/), using the arrow vectors themselves; this allows utilizing a host of DataFrames.jl functionality directly on arrow data; grouping, joining, selecting, etc.
+* `df = copy(DataFrame(Arrow.Table(file)))`: Build a [`DataFrame`](https://juliadata.github.io/DataFrames.jl/stable/), where the columns are regular in-memory vectors (specifically, `Base.Vector`s and/or `PooledVector`s). This requires that you have enough memory to load the entire DataFrame into memory.
```
bkamins, line 69:
```suggestion
* `df = copy(DataFrame(Arrow.Table(file)))`: Build a [`DataFrame`](https://juliadata.github.io/DataFrames.jl/stable/), where the columns are regular in-memory vectors (specifically, `Base.Vector`s and/or `PooledVector`s). This requires that you have enough memory to load the entire `DataFrame` into memory.
```

bkamins (approved):

bkamins:
Thank you!

PR #488, opened 2023-10-14
bump

JoaoAparicio:


baumgold (approved):

Project.toml
```diff
@@ -50,5 +50,5 @@ PooledArrays = "0.5, 1.0"
 SentinelArrays = "1"
 Tables = "1.1"
 TimeZones = "1"
-TranscodingStreams = "0.9.12"
+TranscodingStreams = "0.9, 0.10"
```
ericphanson, line 53: if we need a feature introduced in 0.9.12 (presumably that's why we specified it), we shouldn't relax the compat here:
```suggestion
TranscodingStreams = "0.9.12, 0.10"
```

ericphanson: (and in fact we do, we are relying on a feature introduced in v0.9.12, xref https://github.com/apache/arrow-julia/pull/422)

PR #491, opened 2023-10-16
Fix docs errors

ericphanson:
This package didn't specify `compat` for Documenter, so when the breaking v1 release was made, it broke the docs build (since warnings became errors). This PR fixes those errors and specifies compatibility with Documenter v1.


docs/src/reference.md
```diff
@@ -23,3 +23,11 @@
 Modules = [Arrow]
 Order   = [:type, :function]
 ```
+
+## Internals: `Arrow.FlatBuffers`
```
ericphanson, line 27: documenter wants all docstrings to appear in the docs somewhere. So I've added a section here, but explicitly marked it as part of the internals to not change the API surface.

docs/src/manual.md
```diff
@@ -48,7 +48,7 @@ table = Arrow.Table("data.arrow")
...
...ped", `Arrow.Table` then inspected the metadata in the file to determine the number of columns, their names and types, at which byte offset each column begins in the file data, and even how many "batches" are included in this file (arrow tables may be partitioned into one or more "record batches" each containing portions of the data). Armed with all the appropriate metadata, `Arrow.Table` then created custom array objects ([`Arrow.ArrowVector`](@ref)), which act as "views" into the raw arrow memory bytes. This is a significant point in that no extra memory is allocated for "data" when reading arrow data. This is in contrast to if we wanted to read data from a csv file as columns into Julia structures; we would need to allocate those array structures ourselves, then parse the file, "filling in" each element of the array with the data we parsed from the file. Arrow data, on the other hand, is *already laid out in memory or on disk* in a binary format, and as long as we have the metadata to interpret the raw bytes, we can figure out whether to treat those bytes as a `Vector{Float64}`, etc. A sample of the kinds of arrow array types you might see when deserializing arrow data, include:
```
ericphanson, line 51: Only change here is `ArrowVector` -> `Arrow.ArrowVector` since the reference doesn't work without qualification (as the function isn't exported)

baumgold (approved):

PR #493, opened 2023-12-03
enable field-order-agnostic overloads of `fromarrow` for struct types

jrevels:
Motivated by https://github.com/beacon-biosignals/Legolas.jl/issues/94#issuecomment-1837366852

Still requires:

- [x] docs
- [x] a test
- [x] a bit more due diligence benchmarking-wise. `@benchmark`ing the access in the test case from https://github.com/beacon-biosignals/Legolas.jl/issues/94 didn't reveal any perf difference, which seems like a good sign


src/arraytypes/struct.jl
```diff
@@ -33,23 +33,33 @@ isnamedtuple(T) = false
 istuple(::Type{<:Tuple}) = true
 istuple(T) = false
 
-@propagate_inbounds function Base.getindex(s::Struct{T,S}, i::Integer) where {T,S}
+if isdefined(ArrowTypes, :StructElement)
```
jrevels, line 36: Went with this approach to avoid needing to change Arrow's compat bounds on ArrowTypes

jrevels:
Did some benchmarking, looks like there is a small perf hit here to the general case that needs to be mitigated before this can be merged:

```jl
using Arrow, ArrowTypes, Random, BenchmarkTools

struct Foo
    a::Int
    b::String
    c::Vector{String}
    d::Float64
end

ArrowTypes.arrowname(::Type{Foo}) = Symbol("JuliaLang.Foo")
ArrowTypes.JuliaType(::Val{Symbol("JuliaLang.Foo")}, T) = Foo

genfoo() = Foo(rand(1:10), randstring(10), [randstring(10) for _ in 1:rand(2:5)], rand())
t = (; f = [genfoo() for _ in 1:1000])
f = Arrow.Table(Arrow.tobuffer(t)).f
@benchmark sum(x -> x.d, $f)
```

results on `jrevels:jr/structelement`

```
BenchmarkTools.Trial: 6891 samples with 1 evaluation.
 Range (min … max):  631.291 μs …   2.410 ms  ┊ GC (min … max):  0.00% … 70.30%
 Time  (median):     646.250 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   724.383 μs ± 299.599 μs  ┊ GC (mean ± σ):  10.04% ± 15.20%

  █▅▃▁                                                      ▂▂  ▁
  ████▆▆▅▄▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃███ █
  631 μs        Histogram: log(frequency) by time       1.95 ms <

 Memory estimate: 2.53 MiB, allocs estimate: 22470.
```

results on `main`

```
BenchmarkTools.Trial: 8118 samples with 1 evaluation.
 Range (min … max):  531.667 μs …   2.338 ms  ┊ GC (min … max): 0.00% … 73.71%
 Time  (median):     550.333 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   614.904 μs ± 266.473 μs  ┊ GC (mean ± σ):  9.82% ± 14.78%

  █▇▄▂                                                      ▁▂▁ ▁
  ████▇▆▄▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁███ █
  532 μs        Histogram: log(frequency) by time       1.78 ms <

 Memory estimate: 2.23 MiB, allocs estimate: 18482.
```

jrevels:
Alright, I changed the approach here after profiling different candidate approaches. Looks like leaving `NamedTuple` construction (or even just any intermediate struct construction) in the critical path here borks `getindex` performance, annoyingly, so ended up not going that route. Users of this feature can do still construct `NamedTuple`s themselves, if needed, but this new approach avoids doing so on the default path. 

on this PR now:

```jl
BenchmarkTools.Trial: 8170 samples with 1 evaluation.
 Range (min … max):  530.500 μs …   2.318 ms  ┊ GC (min … max): 0.00% … 74.22%
 Time  (median):     547.833 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   610.966 μs ± 263.948 μs  ┊ GC (mean ± σ):  9.75% ± 14.73%

  █▆▃▁                                                      ▁▃▁ ▁
  ████▇▇▅▃▅▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄███ █
  530 μs        Histogram: log(frequency) by time       1.75 ms <

 Memory estimate: 2.22 MiB, allocs estimate: 18387.
 ```

File: bench.jl [hunk 2]
```diff
@@ -0,0 +1,16 @@
+using Arrow, ArrowTypes, Random, BenchmarkTools
```
ericphanson, line 1: do we want to check this in at top-level? could have a `benchmark` folder

src/ArrowTypes/src/ArrowTypes.jl
```diff
@@ -170,11 +170,15 @@ overload is necessary.
...
-  * `StructKind`: must overload `fromarrow(::Type{T}, x...)` where individual fields are passed as separate
+  * `StructKind`:
+     * May overload `fromarrow(::Type{T}, x...)` where individual fields are passed as separate
      positional arguments; so if my custom type `Interval` has two fields `first` and `last`, then I'd overload like
      `ArrowTypes.fromarrow(::Type{Interval}, first, last) = ...`. Note the default implementation is
      `ArrowTypes.fromarrow(::Type{T}, x...) = T(x...)`, so if your type already accepts all arguments in a constructor
      no additional `fromarrow` method should be necessary (default struct constructors have this behavior).
+     * Alternatively, may overload `fromarrowstruct(::Type{T}, ::Val{fnames}, x...)`, where `fnames` is a tuple of the
+     field names corresponding to the values in `x`. This approach is useful when you need to implement deserialization
+     in a manner that is agnostic to the field order used by the serializer.
```
ericphanson, line 181:
```suggestion
     * Alternatively, may overload `fromarrowstruct(::Type{T}, ::Val{fnames}, x...)`, where `fnames` is a tuple of the
     field names corresponding to the values in `x`. This approach is useful when you need to implement deserialization
     in a manner that is agnostic to the field order used by the serializer. When implemented, `fromarrowstruct` takes precedence over `fromarrow` in `StructKind` deserialization.
```

bench.jl
ericphanson: if we're keeping this, we need to add the license block to the top to pass the CI audit check

ericphanson (approved):
I searched for `Struct{` and found a usage in https://github.com/apache/arrow-julia/blob/787768fc60ff09ea52be0c5fb7c40e04be47619b/src/eltypes.jl#L530, does that need updating?

Generally looks good to me though.

ericphanson (commented):
Oh also, the formatter wants this change, it looks like
```diff
diff --git a/test/runtests.jl b/test/runtests.jl
index 5210685..ed288b3 100644
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1027,7 +1027,11 @@ end
                 end
                 ArrowTypes.arrowname(::Type{Foo493}) = Symbol("JuliaLang.Foo493")
                 ArrowTypes.JuliaType(::Val{Symbol("JuliaLang.Foo493")}, T) = Foo493
-                function ArrowTypes.fromarrowstruct(::Type{Foo493}, ::Val{fnames}, x...) where {fnames}
+                function ArrowTypes.fromarrowstruct(
+                    ::Type{Foo493},
+                    ::Val{fnames},
+                    x...,
+                ) where {fnames}
                     nt = NamedTuple{fnames}(x)
                     return Foo493(nt.x + 1, nt.y + 1)
                 end
```

ericphanson:
@kou would you be able to help initiate a release for v2.7.0?

jrevels:
thanks @ericphanson and @kou !

kou:
Sure!
I'll do it.

kou:
The vote thread: https://lists.apache.org/thread/dxx71lflxtt528rjco8fsjfl255bs628
Please verify and vote the RC!
See also: https://github.com/apache/arrow-julia/tree/main/dev/release#verify

kou:
@ericphanson Could you check my reply on `dev@`? https://lists.apache.org/thread/6ww80f7m16bob09g5g0nx0mj5jyckxnw
I think that we can release a new version once we resolve your problem.

kou:
Passed: https://lists.apache.org/thread/z6y8484wsyd7155cq2jsv4gm4ojpdv3m

ericphanson:
Awesome, and I see you registered both package as well. Thanks very much!

src/ArrowTypes/src/ArrowTypes.jl
```diff
@@ -170,11 +170,15 @@ overload is necessary.
...
   * `ListKind{true}`: for `String` types, they may overload `fromarrow(::Type{T}, ptr::Ptr{UInt8}, len::Int) = ...` to avoid
      materializing a `String`
-  * `StructKind`: must overload `fromarrow(::Type{T}, x...)` where individual fields are passed as separate
+  * `StructKind`:
+     * May overload `fromarrow(::Type{T}, x...)` where individual fields are passed as separate
      positional arguments; so if my custom type `Interval` has two fields `first` and `last`, then I'd overload like
      `ArrowTypes.fromarrow(::Type{Interval}, first, last) = ...`. Note the default implementation is
      `ArrowTypes.fromarrow(::Type{T}, x...) = T(x...)`, so if your type already accepts all arguments in a constructor
      no additional `fromarrow` method should be necessary (default struct constructors have this behavior).
+     * Alternatively, may overload `fromarrowstruct(::Type{T}, ::Val{fnames}, x...)`, where `fnames` is a tuple of the
```
quinnj, line 179: Sorry to be so slow in the review here; it might be worth adding another example the docs here like we have above, something like:
```
So taking the example from before, if my arrow data happens to have the `first` and `last` fields for my `Interval` type reversed, I could implement `fromarrowstruct(::Type{Interval}, ::Val{(last, first}), last, first) = Interval(first, last)`
```
I find concrete examples usually help drive home the core point of a new feature.

PR #500, opened 2024-02-08
migrate ArrowTypes code that depends on Sockets and UUIDs to extensions

baumgold:
Also add support for `Sockets.InetAddr`

### DIFF
--- a/src/ArrowTypes/Project.toml
+++ b/src/ArrowTypes/Project.toml
@@ -18,7 +18,7 @@
 name = "ArrowTypes"
 uuid = "31f734f8-188a-4ce0-8406-c8a06bd891cd"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "2.3.0"
+version = "2.4.0"
 
 [deps]
 
--- a/src/ArrowTypes/Project.toml
+++ b/src/ArrowTypes/Project.toml
@@ -21,8 +21,14 @@ authors = ["quinnj <quinn.jacobd@gmail.com>"]
 version = "2.3.0"
 
 [deps]
+
+[weakdeps]
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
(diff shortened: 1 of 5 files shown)


ericphanson:
What happens if a table was serialized with one of these objects, but now you're loading it in a session in which e.g. UUIDs isn't loaded? I think this is probably breaking, unfortunately, since before you'd get back UUID types, whereas now you have to change your code to load UUIDs to get it.

baumgold:
> What happens if a table was serialized with one of these objects, but now you're loading it in a session in which e.g. UUIDs isn't loaded? I think this is probably breaking, unfortunately, since before you'd get back UUID types, whereas now you have to change your code to load UUIDs to get it.

That's a valid point, thanks for bringing it up.  Do you think it's worth continuing with this change but bump major rather than minor?  Or should we abandon this change for now?

ericphanson:
These are pretty light packages, and are in the sysimage anyway until v1.11, so to me it doesn’t feel worth all the compat bumps it will cause. I think it’s worth waiting until there’s other reasons for a breaking change, or some big reason why having them is causing problems.

baumgold:
Sounds good.  I'll put this PR on hold for now.  Thanks for the feedback.

src/ArrowTypes/Project.toml
```diff
@@ -18,11 +18,17 @@
...
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
+[extensions]
+ArrowTypesSocketsExt = "Sockets"
+ArrowTypesUUIDsExt = "UUIDs"
+
 [compat]
-julia = "1.0"
+julia = "1.9"
```
omus, line 34: Why not keep the existing Julia version compat and do this instead? https://pkgdocs.julialang.org/v1/creating-packages/#Transition-from-normal-dependency-to-extension

PR #504, opened 2024-04-11
Compatibility of schemas with nested types

poncito:
Hi,

Here is a minimal example of the issue I've encountered.

```julia
 struct A
    x::Int
end

struct B
    a::A
end

v = [B(A(i)) for i =1:3]

io = IOBuffer()
Arrow.write(io, v; file=false)
seekstart(io)
Arrow.append(io, v) # throws
```

I don't know if this is really necessary, or if I'm not using this library properly, but this issue makes it difficult to append to arrow files with nested types.

Since I've only added more cases where the call to `append` can succeed, I do not think that this creates retro-compatibility issues.

Thanks for the review!

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1047,12 +1047,12 @@ end
             struct Foo504
                 x::Int
             end
-            
+
             struct Bar504
                 a::Foo504
             end
 
-            v = [Bar504(Foo504(i)) for i =1:3]
+            v = [Bar504(Foo504(i)) for i = 1:3]
             io = IOBuffer()
             Arrow.write(io, v; file=false)
             seekstart(io)
@@ -1061,6 +1061,5 @@ end
             t = Arrow.Table(io)
             @test Arrow.Tables.rowcount(t) == 6
         end
-
     end # @testset "misc"
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1043,16 +1043,16 @@ end
             end
         end
 
-        @testset "# tbd" begin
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -1042,5 +1042,25 @@ end
...
+            v = [Bar504(Foo504(i)) for i =1:3]
+            io = IOBuffer()
+            Arrow.write(io, v; file=false)
+            seekstart(io)
+            Arrow.append(io, v) # testing the compatility between the schema of the arrow Table, and the "schema" of v (using the fallback mechanism of Tables.jl)
+            seekstart(io)
+            t = Arrow.Table(io)
+            @test Arrow.Tables.rowcount(t) == 6
+        end
+
```
baumgold, line 1064: Fix formatting:

```suggestion
            struct Foo504
                x::Int
            end

            struct Bar504
                a::Foo504
            end

            v = [Bar504(Foo504(i)) for I = 1:3]
            io = IOBuffer()
            Arrow.write(io, v; file=false)
            seekstart(io)
            Arrow.append(io, v) # testing the compatility between the schema of the arrow Table, and the "schema" of v (using the fallback mechanism of Tables.jl)
            seekstart(io)
            t = Arrow.Table(io)
            @test Arrow.Tables.rowcount(t) == 6
        end
```

baumgold (changes requested):

baumgold (approved):

poncito:
thanks Ben!

PR #547, opened 2025-04-11
Remove duplicated notifications

raulcd:
Sorry! I did messed up. I did miss the notifications on the top of the file (we have them at the bottom at the rest of the repos) and replicated the section.

This failed to apply the change for discussions as it had incorrect format.


raulcd:
@quinnj sorry, I duplicated this section, I totally missed it was already there. I should have been more careful :)

PR #553, opened 2025-05-10
Add support for running CI on fork

### LINKED ISSUE #552: ci: Add support for running CI on fork
The current configuration runs CI only on the main branch:

https://github.com/apache/arrow-julia/blob/e6ca23797474ac2c5c30198663467d881dc21b2f/.github/workflows/ci.yml#L21-L23

If we want to run CI on fork, we need to push our changes to the main branch instead of a feature branch. It's inconvenient.

kou:
Fixes #552

Push for all branches except branches that are created by Dependabot runs CI. Dependabot creates a branch in apache/arrow-julia and open a PR to apache/arrow-juila. If it happens, CI is triggered by both of push and pull_request. We need only pull_request for the case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,8 +19,11 @@ name: CI
 on:
   pull_request:
   push:
-    branches: [main]
-    tags: ['*']
+    branches:
+      - '**'
+      - '!dependabot/**'
+    tags:
+      - '**'
 jobs:
   license:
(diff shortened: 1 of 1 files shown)


kou:
Triggered on my fork too: https://github.com/kou/arrow-julia/actions/runs/14940472053

quinnj (approved):

PR #568, opened 2025-09-30
add the option to disable the part of threading in `Arrow.Table` that leads to catastrophic negative scaling

KristofferC:
Ref https://github.com/apache/arrow-julia/issues/528.

MWE to see the effect of this PR is:

```julia
using DataFrames, Dates, Arrow, StatsBase, Random, InlineStrings

function generate_data(f)
    number_of_companies = 10000
    dates = collect(Date(2001,1,1):Day(1):Date(2020,12,31))
    companyid = sample(100000:1000000, number_of_companies, replace = false)

    number_of_items = length(companyid)*length(dates)

    df = DataFrame(
            dates = repeat(dates, outer = number_of_companies),
            companyid = repeat(companyid, inner = length(dates)),
            item1 = rand(number_of_items),
            item2 = randn(number_of_items),
            item3 = rand(1:1000,number_of_items),
            item4 = repeat([String7(randstring(['a':'z' 'A':'Z'],5)) for _ in 1:number_of_companies],length(dates))
        )

    @info "Saving to $f"
    open(f, "w") do f
        Arrow.write(f, Tables.partitioner(groupby(df,:dates)))
    end
end

f = joinpath(@__DIR__, "mytestdata.arrow")

if !isfile(f)
    generate_data(f)
end

Arrow.Table(f)
@time Arrow.Table(f)
@time Arrow.Table(f; threaded=false)
```

With the result:

```
  9.964974 seconds (2.57 M allocations: 125.678 MiB, 0.14% gc time, 10610359 lock conflicts)
  0.148734 seconds (2.55 M allocations: 123.376 MiB, 9.03% gc time, 32 lock conflicts, 6.36% compilation time)
```

We can see that disabling this part of the `Table` threading avoids the insane lock contention on https://github.com/JuliaServices/ConcurrentUtilities.jl/blob/5fced8291da84bd081cb2e27d2e16f5bc8081f38/src/synchronizer.jl#L108 and makes the package not grind to a halt when threads are used.


kou:
@quinnj Do you want to include this in the next release? If so, I'll start a release process after this is completed.

quinnj:
I think https://github.com/apache/arrow-julia/pull/570 should fix the original issue; let's review/test/merge that and then cut the release.

Issue #567, opened 2025-09-19
Release 2.8.1


kou:
We need to release a new version before Julia 1.12 release: https://github.com/apache/arrow-julia/pull/543#issuecomment-3310126433

What version should be used for the next version? 2.8.1? (patch release?) 2.9.0? (minor release?)

FYI: We're using semantic versioning.

Could someone check changes since 2.8.0 release?

kou:
Thanks for checking it.

@quinnj @baumgold @ericphanson Do you have any opinion for the next release?

attdona:
News about the next release? My Rembus.jl package fails CI:

```
  93 dependencies successfully precompiled in 297 seconds. 33 already precompiled.
  1 dependency had output during precompilation:
┌ Arrow
│  WARNING: Constructor for type "Type" was extended in `Flatbuf` without explicit qualification or import.
│    NOTE: Assumed "Type" refers to `Base.Type`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function Type end`.
│    Hint: To silence the warning, qualify `Type` as `Base.Type` in the method signature or explicitly `import Base: Type`.
└  
     Testing Running tests...
Package Rembus errored during testing (exit code: 3221225477)
Error: Process completed with exit code 1.
```

kou:
OK. Let's migrate to `pre` from `nightly`. Do you want to open an issue for it?

nhz2:
Just FYI, when running Downstream tests in Julia 1.12 in https://github.com/JuliaIO/CodecZstd.jl/actions/runs/19185940798/job/54852442426, the tests for Arrow.jl seem to stall, but do eventually pass after about 2 hours.

PR #569, opened 2025-10-24
Bump version

kou:
GH-567


quinnj:
I think we need to fix https://github.com/apache/arrow-julia/issues/567#issuecomment-3406664513 first. I think I can look at that tonight.

kou:
OK. (I think https://github.com/apache/arrow-julia/pull/543 fixed it.)

quinnj:
Ah, you are correct! Sorry, I missed that. This sounds good to me then.

Issue #571, opened 2025-10-27
ci: Stable Julia jobs are timed out


kou:
For example: https://github.com/apache/arrow-julia/actions/runs/18793335613/job/53628450445

```text
...
  54 dependencies successfully precompiled in 123 seconds. 27 already precompiled.
     Testing Running tests...
Test Summary: | Pass  Total  Time
ArrowTypes    |  127    127  1.4s
Test Summary:                       | Pass  Total   Time
Date and time wrappers with missing |    4      4  17.1s
Test Summary:      | Pass  Total  Time
`default(T) isa T` |   10     10  0.0s
Error: The operation was canceled.
```

PR #574, opened 2025-10-28
Test on Julia pre instead of Julia nightly

### LINKED ISSUE #573: Default CI should run on Julia `pre` instead of `nightly`
> BTW, do you want to open an issue for the nightly job failure and fix it?
> 
> I'm not sure it makes sense to try to fix something on nightly this early in a Julia release cycle. Also, it looks like several dependencies are broken on the current nightly (TimeZones and Mocking), so I don't know how much we can actually address. It might make more sense to move default CI testing to [`pre`](https://github.com/julia-actions/setup-julia?tab=readme-ov-file#examples), which will do betas, alphas, and RCs when available or otherwise the most recent release. Julia nightly can be quite unstable, and waiting until official alpha/beta will still allow us to test compatibility in a timely manner.  

 _Originally posted by @palday in [#567](https://github.com/apache/arrow-julia/issues/567#issuecomment-3454414884)_

palday:
- fixes #573
- adds a workflow for weekly and manual testing against julia nightly
- adds versioned manifests to the ignore file


.github/workflows/ci_julia_nightly.yml
kou: How about removing `_julia` in file name?
It seems that it's redundant.

iamed2:
The failure seems unrelated to this PR and jobs are getting stuck on main too. Can this be merged?

kou:
OK.

Let's work on the time out problem as #571.


> ```text
> # 126: Test Failed at /home/phillip/Work/arrow-julia/test/runtests.jl:391
>   Expression: tt.a == [1, 2, 3, 1, 2, 3, 4, 1, 2, 3, 4, 5]
>    Evaluated: [1, 2, 3, 1, 2, 3, 4, 1, 2, 3, 4, 4] == [1, 2, 3, 1, 2, 3, 4, 1, 2, 3, 4, 5]
> ```

This is caused by multi-threading:

https://github.com/apache/arrow-julia/actions/runs/19651602551/job/56279594532?pr=574#step:6:219

```text
# 126: Test Failed at /home/runner/work/arrow-julia/arrow-julia/test/runtests.jl:391
  Expression: tt.a == [1, 2, 3, 1, 2, 3, 4, 1, 2, 3, 4, 5]
   Evaluated: [1, 2, 3, 1, 2, 3, 4, 1, 2, 3, 4, 4] == [1, 2, 3, 1, 2, 3, 4, 1, 2, 3, 4, 5]
```

Let's work on it as #420.

PR #594, opened 2026-05-25
Add Arrow C Data Interface support with defensive resource management

robertbuessow:
## Summary

This PR adds support for the [Arrow C Data Interface](https://arrow.apache.org/docs/format/CDataInterface.html), enabling zero-copy data exchange between Julia and C/C++/Python runtimes in the same process.

Key additions and fixes:

- **C Data Interface implementation** (`src/cdatainterface.jl`): `Arrow.from_c_data` / `Arrow.to_c_data` for importing and exporting Arrow arrays and tables across the C ABI boundary. Handles all Arrow types: primitive, boolean, list, fixed-size list, map, struct, union, dict-encoded, and nested types.
- **Defensive resource management**: `CDataHandle` tracks C-side memory lifetime. The GC finalizer uses `jl_safe_printf` instead of `@error` (task switches are forbidden in finalizers) and an atomic counter instead of a non-thread-safe `Ref{Int}`. The finalizer also calls the C release callback as a safety net, so C resources are freed even when `release_c_data` is not called explicitly.
- **ABI compatibility test**: compiles a C probe at test time using `offsetof()` and compares every field offset of `ArrowSchema` and `ArrowArray` against Julia's `fieldoffset()`, catching any struct layout divergence between Julia and the C compiler.
- **Explicit release in tests**: all tests that import C data now call `release_c_data` explicitly, and a final `@testset "no unexpected resource leaks"` asserts `UNRELEASED_HANDLE_COUNT` only increases by the one intentional leak test.
- **Bug fixes**: double-free on import, type narrowing errors, `BoundsError` when dict-encoding `CategoricalArray`s with missing values.

## Test plan

- [ ] `julia --project -e 'using Pkg; Pkg.test()'` passes with no leak warnings
- [ ] `@testset "struct field offsets match C ABI"` — 19 assertions all green
- [ ] `@testset "no unexpected resource leaks"` — counter == initial + 1
- [ ] No `Arrow.CDataHandle GC'd without explicit release_c_data` output during the test run (except inside the `redirect_stderr` block of the intentional-leak test)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -33,7 +33,7 @@ jobs:
         run: which julia
         continue-on-error: true
       - name: Install Julia, but only if it is not already available in the PATH
-        uses: julia-actions/setup-julia@v2
+        uses: julia-actions/setup-julia@v3
         with:
           version: '1'
           arch: ${{ runner.arch }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
           - 2
     steps:
       - uses: actions/checkout@v6
-      - uses: julia-actions/setup-julia@v2
+      - uses: julia-actions/setup-julia@v3
         with:
           version: ${{ matrix.version }}
       - uses: actions/cache@v5
@@ -145,7 +145,7 @@ jobs:
           - x64
     steps:
       - uses: actions/checkout@v6
-      - uses: julia-actions/setup-julia@v2
+      - uses: julia-actions/setup-julia@v3
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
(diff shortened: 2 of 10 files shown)


kou:
Could you read our AI-generated code guideline carefully?
https://arrow.apache.org/docs/dev/developers/overview.html#ai-generated-code

samtalki:
@robertbuessow @kou just wanted to say that we would really appreciate this functionality for PowerIO.jl! https://github.com/eigenergy/PowerIO.jl

Currently implementing a bespoke formulation for the exact same thing and would really appreciate getting this upstreamed. I would be very happy to do a review pass @kou if it could help move this along.

kou:
Could you review this?

You may need to split this PR to small PRs. (I feel that this PR is large to review.)

samtalki:
absolutely @kou ! I will do this as soon as possible. I will try to break this into 2-3 PRs as you prefer and will list @robertbuessow as co author :)

PR #599, opened 2026-06-10
support StringViews 2.0

stevengj:
Updates the package to use StringViews 2.0 if available (which has minor breaking changes with 1.0 that shouldn't affect this package).  In Julia 1.14, where the StringView type is merged into Base, the StringViews 2.0 package becomes a simple stub around the Base type.


quinnj (approved):

kou:
@quinnj Should we release a new version that includes this?

quinnj:
@stevengj, would that be helpful? I typically like to get a batch of changes before a release, unless there's a specific request for one.

stevengj:
I don't think a rapid release is crucial.  The old version will continue to work in the meantime, after all, and I doubt there will be any package in the near future that *requires* the new version.

Issue #184, opened 2021-04-23
Support C data interface


quinnj:
@sa- and I have started work in #178 and #179

Yuan-Ru-Lin:
Hi I wonder what work needs to be done in #178 in order to close this issue. It seems there have been functionalities to load `ArrowSchema` and `ArrowArray` from memory managed by pyarrow already. Do we want to recursively load nested `ArrowSchema` or `ArrowArray`?

ollemartensson:
The original pull request has been chopped up into three distinctive PRs and the one that matters for this issue is now self contained:

#561 

### Changes and updates
- Added a lot of test coverage since the failed CI checks. Property based testing and caught several bugs and actually missing functionality which I think is a testament to how valuable those types of tests can be!
- Updated docs & commit messages & code to follow the ASF Generative Tooling Guidance.

###  ASF Generative Tooling Guidance Assessment
[ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html)
- [x] **Terms and Conditions of Generative AI Tool** 
Claude Code's terms do not place restrictions on use of output that would be inconsistent with the Open Source Definition.
- [x] **Output is not copyrightable subject matter**
- Technical configuration code, standard API implementations, and test patterns are generally not copyrightable
- Implementation follows established Apache Arrow C Data Interface specification
- Code structure follows standard software engineering patterns
- [x] **No third party materials included**
- All code generated specifically for this Apache Arrow.jl implementation
- Based on public Apache Arrow specification, not copied from other implementations
- Original technical design and research contributions
- [x] **Compatible licensing for any referenced materials**
- Apache Arrow C Data Interface specification is publicly available
- All generated code compatible with Apache-2.0 license
- No proprietary or incompatibly licensed content included
- [x] **Disclosure and Attribution Requirements**
- Each commit includes the required disclosure:
     > 🤖 Implementation developed with Claude Code assistance
     >   Research and Technical Design: Original contribution
- This follows the recommended practice of indicating "tooling used to create
  the contribution" and includes it "as a token in the source control commit
  message" as specified in the guidance.
- [x] **Contributor Responsibility**
- Contributor (user) has reviewed and approved all generated content
- Technical design and research contributions are original work
- Implementation methodology developed under direct human guidance
- All content verified for compliance with Apache-2.0 license terms

PR #607, opened 2026-07-12
GH-184: Add Arrow C Data Interface import for null and primitive arrays

samtalki:
## Summary

First PR of the C Data Interface stack for GH-184, extracting null and primitive array import from #603 as requested there.

- Immutable `ArrowSchema` and `ArrowArray` C layout structs with layout assertions.
- `Arrow.from_c_data(schema_ptr, array_ptr; convert=true)` for null and primitive arrays.
- Validates pointer layout, child and buffer counts, offset arithmetic, known flag bits, dictionary pointer consistency, release idempotency, and reads after release. Declared null counts are trusted like Arrow C++ and arrow-rs; `null_count == -1` requires the validity bitmap it is resolved from, like nanoarrow; reserved flag bits are ignored.
- Import moves the base `ArrowSchema` and `ArrowArray` structures into Julia owned storage and marks the sources released without calling their callbacks; callers may free or reuse the passed structures once `from_c_data` returns.
- Aligned buffers are viewed zero copy behind a shared release owner; misaligned fixed width buffers are copied into aligned storage; `copy`, `collect`, `deepcopy`, and `serialize` materialize Julia owned data through the release liveness gate.

Struct, boolean, string, binary, list, and export support land in the follow up PRs.

## Stack

Base is `main`.

Follow up PRs:

- #603 adds struct array import as Tables.jl column tables.
- #604 expands import coverage.
- #605 adds C Data export.
- #606 adds hardening and smoke tests.

## Size

Against `main`: 6 files changed, 1246 insertions, 1 deletion.

## Validation

- `git diff --check main...HEAD`: passed.
- ABI smoke: `ArrowSchema` and `ArrowArray` are isbits structs with expected C layout sizes.
- Focused C Data tests: 93/93 passed.
- Full package tests (bounds checked): 66278/66278 passed.

## AI Assistance

Co-authored-by: Robert Buessow <robert.buessow@relational.ai>
Co-authored-by: Olle Martensson <olle.martensson@gmail.com>
Generated-by: OpenAI Codex


src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+end
+
+const _CDATA_PTR_SIZE = sizeof(Ptr{Cvoid})
+@assert isbitstype(ArrowSchema)
+@assert isbitstype(ArrowArray)
+# On 32 bit ABIs with 8 byte Int64 alignment the C structs contain padding, so
+# the packed size formulas hold only where pointer and Int64 sizes agree.
+@static if Sys.WORD_SIZE == 64
+    @assert sizeof(ArrowSchema) == 7 * _CDATA_PTR_SIZE + 2 * sizeof(Int64)
+    @assert sizeof(ArrowArray) == 5 * _CDATA_PTR_SIZE + 5 * sizeof(Int64)
```
kou, line 49: Does this mean that `sizeof(ArrowSchema)`/`sizeof(ArrowArray)` may be different on 32 bit environment?

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+end
+
+const ARROW_FLAG_DICTIONARY_ORDERED = Int64(1)
+const ARROW_FLAG_NULLABLE = Int64(2)
+const ARROW_FLAG_MAP_KEYS_SORTED = Int64(4)
+
+const _CDATA_MAX_FORMAT_BYTES = 4096
```
kou, line 56: I don't object this for safety but is 4096 reasonable limit?

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+    # instead of waiting. The reentrant acquisition inside release_c_data is
+    # uncontended once trylock succeeds.
+    if trylock(owner.lock)
+        try
+            release_c_data(owner)
+        finally
+            unlock(owner.lock)
+        end
+    else
+        finalizer(_finalize_c_data, owner)
```
kou, line 332: I'm not familiar with Julia but why can we retry the finalizer? If `finalizer()` is called on `owner`, is finalization of the `owner` cancelled?

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+        set = count_ones(@inbounds(bytes[firstbyte + 1]) & firstmask & lastmask)
+    else
+        set = count_ones(@inbounds(bytes[firstbyte + 1]) & firstmask)
+        @inbounds for b = (firstbyte + 2):lastbyte
+            set += count_ones(bytes[b])
+        end
+        set += count_ones(@inbounds(bytes[lastbyte + 1]) & lastmask)
+    end
+    return len - set
+end
```
kou, line 171: Do we need to implement this? (Do we have any existing feature for this?)

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+
+function Base.collect(x::CDataVector{T}) where {T}
+    return _with_live(x) do
+        out = Vector{T}(undef, length(x))
+        for i in eachindex(x)
+            @inbounds out[i] = x[i]
+        end
+        return out
+    end
+end
```
kou, line 198: Why do we need to implement them?

Can users access the the Arrow data after they import the Arrow data?

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+    return stackdict[o] = _detached_owner()
+end
+
+function Base.deepcopy_internal(x::T, stackdict::IdDict) where {T<:CDataVector}
+    haskey(stackdict, x) && return stackdict[x]
+    y = _with_live(x) do
+        T(ntuple(i -> _deepcopy_field(getfield(x, i), stackdict), fieldcount(T))...)
+    end
+    return stackdict[x] = y
+end
```
kou, line 245: Why do we need to implement deep copy?

The Arrow C data interface allows only "moving": https://arrow.apache.org/docs/format/CDataInterface.html#moving-an-array

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+            array.offset,
+            array.n_buffers,
+            array.n_children,
+            array.buffers,
+            array.children,
+            array.dictionary,
+            C_NULL,
+            array.private_data,
+        ),
+    )
```
kou, line 290: I'm not familiar with Julia but we can't update only `array.release` with Julia API, right?

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+        byte == 0x00 && return String(bytes)
+        push!(bytes, byte)
+    end
+    throw(ArgumentError("$name exceeds the import limit"))
+end
+
+function _load_buffers(array::ArrowArray, expected::Int)
+    n_buffers = _checked_nonnegative(array.n_buffers, "ArrowArray.n_buffers")
+    n_buffers == expected ||
+        throw(ArgumentError("ArrowArray.n_buffers does not match the format"))
```
kou, line 411: How about adding the actual and expected N buffers to the error messages for easy to debug?

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+    elseif schema.release == C_NULL || array.release == C_NULL
+        throw(ArgumentError("released Arrow C Data child structure"))
+    end
+    len = _checked_nonnegative(array.length, "ArrowArray.length")
+    offset = _checked_nonnegative(array.offset, "ArrowArray.offset")
+    _checked_add(offset, len, "ArrowArray offset plus length")
+    _checked_nonnegative(array.n_buffers, "ArrowArray.n_buffers")
+    n_children = _checked_nonnegative(array.n_children, "ArrowArray.n_children")
+    schema_children = _checked_nonnegative(schema.n_children, "ArrowSchema.n_children")
+    schema_children == n_children ||
```
kou, line 450:
```suggestion
    schema_n_children = _checked_nonnegative(schema.n_children, "ArrowSchema.n_children")
    schema_n_children == n_children ||
```

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+        throw(ArgumentError("ArrowSchema and ArrowArray child counts differ"))
```
kou, line 451: Can we show actual/expected values in the error message?

src/cdata.jl
```diff
@@ -0,0 +1,647 @@
...
+    null_count = array.null_count
+    if !(null_count == -1 || 0 <= null_count <= len)
+        throw(ArgumentError("ArrowArray.null_count is out of range"))
+    end
+    if (schema.dictionary == C_NULL) != (array.dictionary == C_NULL)
+        throw(ArgumentError("ArrowSchema and ArrowArray dictionary pointers differ"))
```
kou, line 457: Is this error message correct?
It seems that this checks only whether `dictionary` is `NULL` or not.

src/cdata.jl
samtalki: Yes; sizes and padding follow the host C ABI. Clarified this and added a C layout probe in ca63747.

src/cdata.jl
samtalki: I believe so, but I may be wrong. Documented the choice and added boundary tests in ca63747.

src/cdata.jl
samtalki: As I understand, finalizer(f, owner) registers another cleanup attempt, keeping the owner reachable while the current callback returns.

src/cdata.jl
samtalki: I do not believe that this is implemented yet explicitly. The reason it helps is because the existing ValidityBitmap assumes a known null count and a byte-aligned start. This handles null_count == -1 with arbitrary bit offsets. I'm sure there are plenty of other ways to implement it but it's just one choice. Feel free to modify.

src/cdata.jl
samtalki: Yes, exactly. Users can index and iterate the imported array. These are accessor functions that protect buffer reads against concurrent release.

src/cdata.jl
samtalki: Yes, indeed. This effectively allows the user to "exit" the C data interface and construct Julia owned data.

Import moves the C headers. This override handles a _subsequent_ Julia deepcopy(x). It copies the buffers while preventing duplication of producer release callbacks.

My personal opinion is that it would be somewhat user hostile to not include this. However, I could be totally wrong. If you think that this doesn't belong in this package, please feel free to remove it.

src/cdata.jl
samtalki: I think so. The loaded Julia value is immutable. A field-pointer store is possible; this helper writes a replacement header while preserving every other field.

src/cdata.jl
samtalki: Agreed. Added actual and expected counts in ca63747.

src/cdata.jl
samtalki: Applied in ca63747 and carried through the restack.

src/cdata.jl
samtalki: Added both counts and their source structures in ca63747.

src/cdata.jl
samtalki: Corrected in ca63747: ArrowSchema.dictionary and ArrowArray.dictionary must both be NULL or both be non-NULL.

samtalki:
Resolved and restacked. Thanks for your patience @kou

quinnj:
@samtalki thanks for all the work on this stack, including splitting it up for review and working through the ownership, release, and validation details with @kou.
I want to flag the overlap with #609 before moving ahead. I’m planning to merge that rewrite for Arrow.jl 3.0. It changes the underlying array and ownership model and includes C Data and C Stream support, so I expect that to be the implementation we build on.
I realize you’ve put substantial time into getting this ready against the current code, and I apologize for the lack of communication until now on parallel efforts (I've been playing around w/ the major rewrite for a while now, but leading up to JuliaCon finally found the time to actually implement). 

You mentioned needing this for PowerIO.jl. If you’re interested, I’d welcome your feedback on whether the new C Data API covers that use case. There’s no expectation that you rework this whole stack against the new internals.

My goal w/ the rewrite though was to make the C data model the core of everything since it's the more foundational layer, instead of building it on top of the, IMO, now dated architecture of the original Arrow.jl. It allows us to easily build the C data, stream, IPC, file formats, and normal Julia interop (Tables.jl, etc.) as small layers/adapters on top.

PR #609, opened 2026-08-20
Rewrite Arrow.jl internals for 3.0

### LINKED ISSUE #610: `Arrow.Table` uses last dictionary for all batches in streaming decode, resulting in incorrect reads
This issue was primarily written by AI, but checked by a person. The practical upshot is that data is read back differently than it was written.

Example: 

Here's a 2-batch stream where the second `DictionaryBatch` replaces (not deltas) the dictionary for column a: batch 1 uses ["alpha","beta"], batch 2 uses ["gamma","delta"]. Reading it back with `Arrow.Table` gives `["gamma","delta","gamma","delta"]` instead of `["alpha","beta","gamma","delta"]`; `Arrow.Stream` on the same bytes gives the correct answer.

```julia
using Arrow

# Two independent single-batch streams, same schema, different dict pools.
# Each is written fresh, so each DictionaryBatch is "initial" (isDelta=false) —
# exactly what a producer like pyarrow emits when per-batch pools differ.
open(io -> Arrow.write(io, (a=Arrow.DictEncode(["alpha", "beta"]),)), "s1.arrows", "w")
open(io -> Arrow.write(io, (a=Arrow.DictEncode(["gamma", "delta"]),)), "s2.arrows", "w")

# Splice: [schema, dict1, rb1] from s1 + [dict2, rb2] from s2 + EOS.
# This reproduces a foreign stream whose 2nd DictionaryBatch replaces dict id 0.
function messages(bytes)
    spans, pos = UnitRange{Int}[], 1
    while pos + 3 <= length(bytes) && Arrow.readbuffer(bytes, pos, UInt32) == Arrow.CONTINUATION_INDICATOR_BYTES
        pos += 4
        msglen = Arrow.readbuffer(bytes, pos, Int32)
        msglen == 0 && break
        pos += 4
        msg = Arrow.FlatBuffers.getrootas(Arrow.Meta.Message, bytes, pos - 1)
        start = pos - 8
        pos += msglen + msg.bodyLength
        push!(spans, start:(pos - 1))
    end
    spans
end

b1, b2 = read("s1.arrows"), read("s2.arrows")
s1, s2 = messages(b1), messages(b2)
eos = reinterpret(UInt8, [Arrow.CONTINUATION_INDICATOR_BYTES, UInt32(0)])
write("combined.arrows", vcat(b1[s1[1]], b1[s1[2]], b1[s1[3]], b2[s2[2]], b2[s2[3]], eos))

println(collect(Arrow.Table("combined.arrows").a))                    # ["gamma","delta","gamma","delta"] -- WRONG
println([collect(b.a) for b in Arrow.Stream("combined.arrows")])      # [["alpha","beta"],["gamma","delta"]] -- correct
```

The `pyarrow` version behaves correctly in this case:

```python
import pyarrow as pa, pyarrow.ipc as ipc
b1 = pa.record_batch({'a': pa.DictionaryArray.from_arrays(
        pa.array([0, 1], pa.int32()), pa.array(['alpha', 'beta']))})
b2 = pa.record_batch({'a': pa.DictionaryArray.from_arrays(
        pa.array([0, 1], pa.int32()), pa.array(['gamma', 'delta']))})
with ipc.new_stream('r2.arrows', b1.schema) as w:
    w.write_batch(b1); w.write_batch(b2)

ipc.open_stream('r2.arrows').read_all().column(0).to_pylist()
# ['alpha', 'beta', 'gamma', 'delta']
```

quinnj:
## Summary

- preserve per-batch dictionary snapshots, accept undeclared stream replacements, and read dictionary deltas in streams, files, and range scans; fixes #610
- replace the Arrow 2.x internals with a validated core, new IPC reader and writer, C Data and C Stream support, and a materialized public table facade
- compile each `Tables.Scan` once, keep storage and public-domain plans separate, preserve stable empty-result schemas, and support sparse byte-range reads with footer statistics
- deepen column construction around one recursive adapter for native, retained-schema, dictionary, Union, and ArrowTypes-backed columns
- retain automatic ArrowTypes custom serialization, public-domain lifting, extension metadata, and nested custom-type round trips
- use registered DataStrings 1.0 as the StringView representation package and keep Arrow.jl's export surface small
- use DataDecimals 1.0 and Durations 1.0 for supported top-level decimal and interval columns; preserve wire descriptors on rewrite
- discard inherited batch statistics when rewriting edited tables so filters cannot skip changed rows
- reject invalid writer options before opening destination files, release owned files after failures, and preserve the original error
- save only referenced Utf8View/BinaryView bytes, retain shared ranges, and avoid sorting and per-entry copies when referenced bytes already cover each copied buffer prefix
- remove JSON3 and StructTypes integration and support claims
- deepen deterministic fuzzing, Apache Arrow corpus coverage, PyArrow/nanoarrow oracles, C Data checks, allocation limits, trim checks, documentation, and package administration

## Dictionary update validation

- final head `51ccbcc33cd145c73bd7d43a2dcb2bc19935e671`: full Julia 1.12.6 package suite passes with bounds checks, CloudStore integration, all four acceptance batteries, and Aqua
- final head passes all 55/55 GitHub checks; push CI, pull-request CI, and Conformance all completed successfully, including pinned JuliaFormatter verification
- 474/474 focused assertions pass on Julia 1.10.11 and 1.12.6 with bounds checks enabled; the final test-helper reuse path also passes all 474 assertions
- frozen PyArrow 25.0.1 fixtures cover undeclared replacements, dictionary deltas, V4/V5 messages, file/stream input, null values, nested pools, and uncompressed/LZ4/Zstd bodies
- PyArrow fully validates 33 generated delta streams and confirms exact Arrow value equality for both batches
- delta concatenation preserves physical Union routes and earlier dictionary snapshots, charges the reader allocation budget, and rejects missing bases, invalid indices, malformed payloads, and full replacements in files
- V4 messages with PyArrow's V5 file footer are accepted; message-version changes remain errors
- reader support shares dictionary transitions across stream, file, and range reads; writer dictionary emission is unchanged

## Dependency

JuliaData/Tables.jl#380 is merged and released as Tables.jl 1.14.0. Commit dd8c272 removes the temporary source and CI pins, raises the Tables.jl compatibility floor to 1.14, and resolves Tables from the General registry everywhere (package, CI workflows, conformance image, docs).

## Breaking changes

This is the Arrow.jl 3.0 rewrite. Important changes include:

- Julia 1.10 is now required.
- `Arrow.Table` and `Arrow.Stream` return materialized Julia vectors instead of lazy Arrow vector wrappers.
- `Arrow.write` is eager and whole-buffer. `Arrow.Writer` and IPC-stream append are reimplemented with fixed-schema semantics; the curried write and `tobuffer` compatibility forms remain. Obsolete writer tuning keywords are ignored with warnings.
- `Arrow.write(io, table)` writes file format by default; use `file=false` for stream format.
- ArrowTypes-based custom serialization remains automatic. `ArrowTypes` stays exported only as an Arrow 2.x compatibility exception. Packages that define mappings should depend on and `import ArrowTypes` directly. `Arrow.ArrowTypes` remains as a qualified compatibility binding.
- most APIs are no longer exported and must be called through `Arrow`.

Fresh declared heterogeneous Julia Union columns are written as canonical dense Arrow Unions. A retained Union rewrite fails closed when materialization has lost its original route. Nested retained dictionary pools follow the same fail-closed rule.

See `CHANGELOG.md` and `docs/src/migration.md` for the complete list and migration guidance.

## Shared-value validation

- Full local package suite passed on Julia 1.12.7, including CloudStore integration and Aqua.
- ArrowCore and registered DataStrings trim compilation and execution passed.
- Two-way PyArrow 25.0.1 checks passed for all four decimal precision tiers, including exact values, precision, scale, and nulls.

## Validation before shared-value adoption

- prior head: `135bac33831e09ee994f6912fa0312c22497158b`
- Julia 1.10.11: 2,401/2,401 assertions across view output, rewrite, writer failure, incremental writer, facade, and combined-behavior tests, with bounds checks enabled
- Julia 1.12.6: 770/770 view-output assertions and 69/69 additional checks combining edited rows, saved text, file preservation, and writer failures
- PyArrow 25.0.1: all six file/stream and compression combinations pass full validation; both Python-to-Julia-to-Python cases pass
- strict Documenter build and pinned JuliaFormatter 2.12.4 check pass
- full Julia 1.12.6 suite: 8,057/8,057 assertions, all four acceptance batteries, and Aqua pass in one normal invocation with bounds checks enabled
- prior head `135bac3` passes 59/59 GitHub checks across push CI, pull-request CI, and Conformance; all three workflows completed successfully

## Earlier rewrite validation

- the previous head `67e17cf` passes 115/115 GitHub checks across the push, pull-request, and Conformance workflows, including Windows Julia latest
- the earlier head `9878699` passes 59/59 GitHub checks across the push, pull-request, and Conformance workflows
- the earlier head `9878699` passes the complete root package suite in a fresh Julia 1.12.7 environment with registered Tables 1.14.0
- the earlier head `9878699` is clean under the exact CI formatter: Julia 1.12.7 and JuliaFormatter 2.12.4
- focused final checks pass: ArrowStrings 1.0 has 2,537/2,537 passing assertions; acceptance and conformance support has 216/216; Documenter builds from a clean environment; `git diff --check` and release-script syntax checks pass
- the Tables.jl 1.14 registry switch passes the full root suites on Julia 1.10.11 and 1.12.6 against registered Tables 1.14.0, the JuliaC trim gate 6/6, Documenter, JuliaFormatter 2.12.4, and all three docker conformance suites (corpus, oracle, cdata) in the rebuilt image
- all four acceptance batteries pass on both supported test versions: 159/159 named checks, including 47 Scan checks and real statistics-pruned no-fetch proofs
- seeded properties pass 1,369/1,369 on both supported test versions
- minimum ArrowTypes 2.0 compatibility passes 7/7 on Julia 1.10; the full ArrowTypes suite includes 49/49 public-domain Scan assertions
- deterministic extended fuzzing passes 512 cases, 3,072 variants, 3,072 rewrites, 27,648 Scan checks, 49 layout checks, and 20,000 malformed mutations across all 41 routes with 119 determinism replays
- malformed mutation outcomes: 561 accepted by the format, 19,439 clean validation failures, 0 unexpected allocation-limit classifications, and no raw exception leaks
- Apache Arrow corpus: 388 pass, 0 fail, 63 declared skips
- PyArrow/nanoarrow oracle: 315 pass, 0 fail, 20 declared skips
- C Data oracle: 147 pass, 0 fail, 8 declared skips
- JuliaC trim-safe compile passes 6/6 checks with zero verifier errors and a successful generated binary
- independent standards and specification reviews report CLEAN with no strong or worthwhile findings

## Release preparation

- DataStrings, DataDecimals, and Durations 1.0 resolve from General; the bundled ArrowStrings package and its release steps are removed
- the Apache release-candidate template and CI now enforce a vote window of at least 72 hours
- Arrow.jl remains at `3.0.0-DEV` pending the rewrite merge and Apache release process

Co-authored by Codex


davidanthoff:
Very cool!

One question: could ArrowCore.jl be a real standalone, dependency-free package, similar to how ArrowStrings.jl is split out already? From reading the code, it looks like ArrowCore.jl could contain what it has at the moment in this branch, plus cdata.jl, and the uncompressed IPC read/write path (with the vendored FlatBuffers/Meta layer).  And the Tables.jl facade, ArrowTypes lowering, and the compression codecs could then stay in Arrow.jl on top of it.

As far as I can tell this split already exists at the code level: ArrowCore only imports Mmap, cdata.jl only uses ArrowCore, and neither ipc_read.jl nor ipc_write.jl references Tables at all — the codecs only appear at the buffer (de)compression step. So ArrowCore.jl would depend on Base and stdlibs only: no binary deps (Lz4_jll/Zstd_jll), no Tables.jl. The two places where the package boundary doesn't quite match the code today seem small: EnumX (but Meta is generated, so the generator could emit self-contained enum modules instead), and compression (which could be a plain dispatch seam — ArrowCore.jl defines compress/decompress generics and errors helpfully on compressed buffers, Arrow.jl keeps its hard codec deps and implements them; so using Arrow behaves exactly as it does now.

The reason I'm asking: we could use this in the Julia VS Code extension. The REPL process could use ArrowCore.jl and ArrowStrings.jl to hand large tables to the extension's interactive table viewer — write an uncompressed IPC file, and the TypeScript side reads it with Arrow JS, with the file-format footer giving the viewer random access to just the visible row range. That would be really nice! But the way we ship code that loads into the user's REPL process rules out anything with a binary dependency. And we also couldn't use anything that pulls in Tables.jl, because that would pin the Tables.jl version users see in their REPL to whatever we ship. So this could only work if there is a core package without those deps.

(I'd also have some uses for this in Queryverse, but the VS Code one probably the more interesting one)

And, as all ideas these days, heavily Claude helped ;)

quinnj:
I think it's close to that, but currently we still have the new ArrowStrings.jl package which provides the native offset + string view representation. We want that as a stand-alone package because I want to be able to use it as a dep for CSV.jl to do direct parse into that representation, I.e. we can parse a csv file directly into a zero-copy arrow representation.

So not quite zero-dependency.

I'm also not sure just ArrowCore.jl would be quite enough to have arrow JS be able to read easily. Currently we have the CData & IPC layers on top of the core buffers to actually fulfill those interfaces.

I do have another project/data format I've been working on that would be a much cleaner/simpler binary table format with cross-language implementations. I've been slowly iterating on it, but it could potentially be a really good fit for this.

davidanthoff:
Yeah, I also think it is super close.

I think the ArrowStrings.jl would not be a problem, i.e. if ArrowCore.jl were to depend on ArrowStrings.jl that would work (we use a trick for that kind of situation a lot in the extension). The main thing is really that the compression stuff would have to not be in there, because we can't handle the binary dependencies that are pulled in.

And yes, the cdata and ipc layers would have to move into ArrowCore.jl as well. Claude thinks that would be easy, but who knows ;)

